PR #1, opened 2021-08-10
Initial Commit

raycar5:


README.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. Set the chain endpoint in [project.yaml](project.yaml).network.endpoint.
+2. Set the starting block in [project.yaml](project.yaml).dataSources.main.startingBlock (this repo has been designed to work from the start of spec version 2021, aka block 1187881)
+3. Install subql cli: `npm i -g @subql/cli`
+4. `rerun.sh` (requires docker compose)
+
+## Comparing output with harvester
+
+1. Set the right configuration in `scripts/.env`
+2. `npm run compare`
```
polymath-eric, line 15: (nit) should this be `yarn run compare` ? Functionally should be the same, but this repo does have a `yarn.lock`.

README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Comparing output with harvester
+
+1. Set the right configuration in `scripts/.env`
+2. `npm run compare` 
+
+## Using event_arg_x indexes.
+The event_arg_x columns are now text, in order for them to fit in BTree indexes they are truncated to 100 characters like in the harvester.
+This means that if you want to take advantage of the index in your query, you must use an expression like: `WHERE left(event_arg_x, 100) = 'foobar'`.
+Otherwise using `WHERE event_arg_x = 'foobar'` will result in a full table scan.
```
polymath-eric, line 20: This seems like its going to cause full table scans, eventually someone is going to forget to use `left`.

I tried out a GIN index and that seems to get hit without needing to use `left`

```sql
create extension btree_gin; -- (not sure why extension needs to be made)
create index gin_arg_1 on events using gin(event_arg_1)
```
then on a query I get this as an explain

```sql
explain select * from events e 
WHERE event_arg_1 = 'foobar'
```

```
Bitmap Heap Scan on events e  (cost=20.01..24.02 rows=1 width=650)
  Recheck Cond: (event_arg_1 = 'foobar'::text)
  ->  Bitmap Index Scan on gin_arg_1  (cost=0.00..20.01 rows=1 width=0)
        Index Cond: (event_arg_1 = 'foobar'::text)
```

Maybe we should use a GIN index instead? I am not sure on the use case here, but needing to call `left` for every where clause doesn't seem good to me. Not sure if there is another way to solve this, or if its even an issue though.

polymath-eric: I talked to Diego about this.

For now we are thinking of leaving this as is, but in the future we can investigate more if it turns out to be an issue.

README.md
raycar5: fixed

project.yaml
```diff
@@ -0,0 +1,4805 @@
+specVersion: 0.0.1
+description: ""
+repository: ""
+schema: ./schema.graphql
+network:
+  endpoint: wss://itn-rpc.polymesh.live
```
satyamakgec, line 6: Do we want this URL environment-based? or different branches may have this different URL.

project.yaml
```diff
@@ -0,0 +1,4805 @@
...
+                    {
+                      "primary_key": "AccountId",
+                      "secondary_keys": "Vec<SecondaryKey>",
+                    },
+                  "KeyIdentityData":
+                    {
+                      "identity": "IdentityId",
+                      "permissions": "Option<Permissions>",
+                    },
+                  "CountryCode":
```
satyamakgec, line 3813: Why it is defined multiple times ?

scripts/compare.ts
```diff
@@ -0,0 +1,401 @@
...
+  1845838, 1845890, 1845933, 1845934, 1845952, 1846013, 1846024, 1846095,
+  1846135, 1846209, 1846268, 1846383, 1846407, 1846421, 1846434, 1846440,
+  1846442, 1846485, 1846538, 1846568, 1846615, 1846626, 1846654, 1846723,
+  1846725, 1846736, 1846785, 1846792, 1846798, 1846828, 1846911, 1846916,
+  1846920, 1846954, 1847267, 1851770, 1853306, 1856560, 1857874, 1858843,
+  1862076, 1866430, 1867999, 1870625, 1871689, 1872721, 1875733, 1876039,
+  1877112, 1877994, 1878136, 1880942, 1887033, 1891663, 1891667, 1891681,
+  1891703, 1891779, 1891991, 1892061, 1892111, 1892151, 1892152, 1892181,
+  1892563, 1893416, 1894739, 1895907, 1898781, 1900291,
+]);
```
satyamakgec, line 101: For the mainnet we don't need this kind of work? How we can make this code mainnet ready?

scripts/compare.ts
```diff
@@ -0,0 +1,401 @@
...
+      throw new Error(`Unknown table: ${table}`);
+  }
+};
+
+main()
+  .then(() => process.exit(0))
+  .catch((e) => {
+    console.log(e);
+    process.exit(1);
+  });
```
satyamakgec, line 401: Again we don't need this for the mainnet ?

satyamakgec:
Why does the test and linting are failing @raycar5 ?

project.yaml
raycar5: I take the `polymesh_schema`.json from each version and add it directly, diffing it would be error-prone

scripts/compare.ts
raycar5: no, they are wrong blocks in the harvester

raycar5:
Because I had a small mistake in some previous changes, it's fixed now

project.yaml
raycar5: yes, It's something I'm still thinking how to solve exactly

project.yaml
satyamakgec: Okay

PR #12, opened 2021-09-30
Add Funding event handler

polymath-eric:


src/mappings/entities/fundingEvent.ts
```diff
@@ -0,0 +1,29 @@
...
+export async function handleFundingEvent(
+  block_id: number,
+  event_id: string,
+  module_id: string,
+  params: Codec[],
+  event: SubstrateEvent
+): Promise<void> {
+  if (module_id === "asset" && event_id === "Issued") {
+    const rawTicker = params[1];
+    const ticker = hexToAscii(rawTicker.toString());
```
raycar5, line 15: the way I do it is: `removeNullChars(u8aToString(item.toU8a()))` could you perhaps create a util function serializeTicker that is shared here and in `serializeLikeHarvester.ts` please?

raycar5 (approved):

PR #14, opened 2021-10-04
Rename event handlers to the entity they are mapping

polymath-eric:


src/mappings/entities/mapExternalAgent.ts
```diff
@@ -4,7 +4,10 @@ import { TickerExternalAgentAction, Sto } from "../../types";
 import { getOrDefault, serializeTicker } from "../util";
 import { EventIdEnum, ModuleIdEnum } from "./common";
 
-export async function handleTickerExternalAgentAction(
```
raycar5, line 7: The entity is a TickerExternalAgentAction

src/mappings/entities/mapStaking.ts
```diff
@@ -19,8 +19,10 @@ const bondedUnbondedOrReward = new Set([
   StakingEventType.Unbonded,
   StakingEventType.Reward,
 ]);
-
-export async function handleStakingEvent(
+/**
+ * Subscribes to events related to staking events
+ */
+export async function mapStaking(
```
raycar5, line 25: The entity is a StakingEvent

raycar5 (changes requested):
I'm fine with changing it from `handle` to `map` but my mappers were already named after the entities they represent, not the event.

Also don't forget to rename the Funding entity in the` schema.graphql`

src/mappings/entities/mapStaking.ts
polymath-eric: My bad, when I was doing funding I saw these and though I was supposed to add Event to the name.

Had another look at tooling and see my mistake now.

raycar5 (approved):

PR #15, opened 2021-10-05
External agents mappers

raycar5:


raycar5:
I am very dumb

src/mappings/entities/mapTickerExternalAgentHistory.ts
```diff
@@ -0,0 +1,291 @@
...
+  if (
+    moduleId === ModuleIdEnum.Externalagents &&
+    eventId === EventIdEnum.AgentRemoved
+  ) {
+    const did = params[2].toString();
+    const ticker = serializeTicker(params[1]);
+
+    const promises = [
+      removeMember(did, ticker),
+      (async () => {
```
polymath-eric, line 169: (nit/ignorable) I don't think we need an anonymous function here since nothing happens after await.

polymath-eric (approved):

PR #16, opened 2021-10-05
NCBD-460 Authorization handler

prashantasdeveloper:
### JIRA Link 

https://polymath.atlassian.net/browse/NCBD-460

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -102,3 +102,16 @@ type Sto @entity {
   id: ID!
   offering_asset: String!
 }
+
+type Authorization @entity {
+  id: ID!
+  block_id: Int!
+  auth_id: Int! @index
+  type: String @index
+  from_did: String @index
+  to_did: String @index
+  to_key: String @index
+  data: String
+  expiry: String
+  status: String!
+}
--- a/src/mappings/entities/authorization.ts
+++ b/src/mappings/entities/authorization.ts
@@ -1,0 +1,78 @@
+import { Codec } from "@polkadot/types/types";
+import { Authorization } from "../../types";
+import {
+  getFirstKeyFromJson,
+  getFirstValueFromJson,
+  getTextValue,
+} from "../util";
+import { EventIdEnum, ModuleIdEnum } from "./common";
+
+enum AuthorizationEventType {
+  AuthorizationAdded = "AuthorizationAdded",
+  AuthorizationRevoked = "AuthorizationRevoked",
+  AuthorizationRejected = "AuthorizationRejected",
+  AuthorizationConsumed = "AuthorizationConsumed",
+}
+
+enum AuthorizationStatus {
+  Pending = "Pending",
+  Consumed = "Consumed",
+  Rejected = "Rejected",
+  Revoked = "Revoked",
+  Expired = "Expired",
+}
+
+const authorizationEventTypes = new Set<string>(
+  Object.values(AuthorizationEventType)
+);
+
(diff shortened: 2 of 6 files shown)


src/mappings/entities/authorization.ts
```diff
@@ -0,0 +1,78 @@
...
+import { Authorization } from "../../types";
+import {
+  getFirstKeyFromJson,
+  getFirstValueFromJson,
+  getTextValue,
+} from "../util";
+import { EventIdEnum, ModuleIdEnum } from "./common";
+
+enum AuthorizationEventType {
+  AuthorizationAdded = "AuthorizationAdded",
```
raycar5, line 11:
```suggestion
  AuthorizationAdded = EventIdEnum.AuthorizationAdded,
```

src/mappings/entities/authorization.ts
```diff
@@ -0,0 +1,78 @@
...
+        block_id,
+        auth_id: params[2],
+        to_key: getTextValue(params[1]),
+        status: authorizationEventStatusMapping.get(event_id),
+      }).save();
+    } else {
+      const expiry = getTextValue(params[5]);
+
+      let status = AuthorizationStatus.Expired;
+      if (!expiry || +expiry > new Date().getTime()) {
```
raycar5, line 60: As discussed in slack we can't process the expired status since this code runs once and never updates, therefore if someone queries the entity hours from now, it might be expired by then, the best we can do is allow people to filter out expired authentications.

This means that if a consumer depends on this status enum, they must now calculate if the status is expired on the client side

raycar5 (changes requested):

raycar5 (changes requested):

schema.graphql
```diff
@@ -102,3 +102,16 @@ type Sto @entity {
   id: ID!
   offering_asset: String!
 }
+
+type Authorization @entity {
+  id: ID!
+  block_id: Int!
```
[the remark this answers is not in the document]
prashantasdeveloper: so in this case, should we also keep an updated block (for referencing the block when status is updated?)

src/mappings/entities/authorization.ts
prashantasdeveloper: Removed the code and status.

prashantasdeveloper: Should we keep a granular level change-log (where we are either discontinuing something or changed the logic quite considerably) somewhere that can be useful for SDK/anybody consuming graphql directly later on ?

cc. @satyamakgec @raycar5 @polymath-eric

raycar5: I think we will have to write a `Migration guide` for consumers indeed

raycar5 (changes requested):

schema.graphql
```diff
@@ -102,3 +102,17 @@ type Sto @entity {
...
+  created_block: Int
+  auth_id: Int! @index
+  type: String @index
+  from_did: String @index
+  to_did: String @index
+  to_key: String @index
+  data: String
+  expiry: Date @index
+  status: String!
+  updated_block: Int
```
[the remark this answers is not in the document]
prashantasdeveloper: data was not mandatory field as specified in tooling schema, so leaving it optional

src/mappings/entities/authorization.ts
polymath-eric: I think we should have a doc where we can record these things as we go along, even if its in bullet point form, so we can go back later.

PR #18, opened 2021-10-07
NCBD-465 Migrate Investment action

prashantasdeveloper:
JIRA Link - https://polymath.atlassian.net/browse/NCBD-465


schema.graphql
```diff
@@ -116,3 +116,15 @@ type Authorization @entity {
   status: String!
   updated_block: Int!
 }
+
+type Investment @entity {
+  id: ID!
+  block_id: Int!
+  investor: String!
+  sto_id: Int!
+  offering_token: String!
```
polymath-eric, line 125: Looking in tooling it looks like users can search for this by passing in the "ticker" param. (event_arg_2, double check please).

I think we should `@index` `sto_id` + `ticker`. and document `ticker` -> `offering_ticker` change of params. 

I don't think we need to worry about a composite index, I think a (ticker, sto_id) query would use the ticker index and be able to scan the few rows it gets from that.

polymath-eric (approved):
LGTM

raycar5 (approved):

PR #26, opened 2021-10-18
Add tokens held mapper

polymath-eric:
Adds support for keeping track of all tokens ever held by DID.

I did the approach of making an entity, but it makes the query a bit awkward since there is more nesting and it introduces extra methods in the schema that aren't really useful,  e.g. `tokensHelds { ...`

I am thinking it might be cleaner to make an identity entity and use a JSON field to store an array so the query can be more natural

```
query{
  identity(id: "0x050") {
    tokensHeld
  }
}
```

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -237,3 +237,9 @@ type Proposal @entity {
   votes: [ProposalVote]! @derivedFrom(field: "proposal")
   snapshotted: Boolean!
 }
+
+type TokensHeld @entity {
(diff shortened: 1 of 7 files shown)


raycar5:
The question is, will there be any identity with so many tokens held that it requires pagination, if yes then we should use this approach, if not then I agree the json field could be acceptable

raycar5 (approved):
Name change double thumbs up

PR #39, opened 2021-11-16
feat: 🎸 Wrap up subquery native graphql

raycar5:
Added trustedClaimIssuerTicker entity and tweaked identitiesWithClaims


src/mappings/mappingHandlers.ts
```diff
@@ -147,6 +149,7 @@ export async function handleEvent(event: SubstrateEvent): Promise<void> {
   await Event.create({
     id: `${blockId}/${eventIdx}`,
     blockId,
+    parentBlockId: `${blockId}`,
```
polymath-eric, line 152: Why do we need this when its the same as the blockId?

schema.graphql
```diff
@@ -1,15 +1,16 @@
 type Event @entity {
   id: ID!
   blockId: Int!
+  parentBlock: Block!
```
polymath-eric, line 4: Why not change `blockId: Int!` -> `block: Block!` instead? 

That should keep the `blockId` field as is for clients + provide the block connection and shouldn't even need a code change.
 
 Either way the connection should have a test case somewhere.

src/mappings/entities/mapTrustedClaimIssuerTicker.ts
```diff
@@ -0,0 +1,33 @@
...
+import { TrustedClaimIssuerTicker } from '../../types/models/TrustedClaimIssuerTicker';
+import { serializeTicker } from '../util';
+import { EventIdEnum, ModuleIdEnum } from './common';
+
+export async function mapTrustedClaimIssuerTicker(
+  _: number,
+  eventId: EventIdEnum,
+  moduleId: ModuleIdEnum,
+  params: Codec[],
+  __: SubstrateEvent
```
polymath-eric, line 12: (nit) Pretty annoying eslint doesn't handle this though.

```suggestion
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
  __: SubstrateEvent
```

?

polymath-eric (changes requested):

src/mappings/mappingHandlers.ts
prashantasdeveloper: Referring to PR https://github.com/PolymathNetwork/polymesh-sdk/pull/599/files, in order to map the response from the new middleware as the old one, it was required to add parentBlock. (In sdk while fetching events the gql selected fields includes `parentBlock { datetime }`) 

Seems more of a hack.

schema.graphql
prashantasdeveloper: I like idea, but then we would have to again map for each query to match the current o/p (again saying based on current implementation because of the [PR](https://github.com/PolymathNetwork/polymesh-sdk/pull/599/files)). I think originally the idea was to create keep the structure as close as the current one and in doing so it was kept as `blockId: Int!`.

polymath-eric: When SubQuery sees the relation it generates the id column, and makes the link. We should still have a `blockId` field even after the rename so the schema doesn't change.

prashantasdeveloper:
This will be abandoned in favour of https://github.com/PolymathNetwork/polymesh-subquery/pull/54

PR #43, opened 2022-01-12
feat: Add asset entity, map asset

honarpour:
This PR:
- [x] Add call ID enums: [`CallIdEnum`](https://github.com/PolymathNetwork/polymesh-subquery/pull/43/files#diff-838e0284467b77133abd8ca58e3dcf6e4458fe320a97a67bb993e6ab6376f8faR359)
- [x] Add asset entity to GQL schema: [`Asset`](https://github.com/PolymathNetwork/polymesh-subquery/pull/43/files#diff-efc1675187620595b0844197a25c35eac9b6df752f9182e5cffddee6f27a8489R97)
- [x] Add asset map util to extrinsic handler to update basic token info: [`handleCall`](https://github.com/PolymathNetwork/polymesh-subquery/pull/43/files#diff-cacfd1ec0155919c9a4ad17775cce37b7d92a736f89d84216e087d8f19405385R162), [`mapAsset`](https://github.com/PolymathNetwork/polymesh-subquery/pull/43/files#diff-b06eb3079d9096a941e7e4d5b51e272d9665364d683edf2324d73bfff50329b4R226)
- [x] Handle updating asset holders and total transfers (settlements)
- [x] Handle asset ownership transfers
- [x] Handle compliance rules
- [ ] Handle ownership limits


schema.graphql
```diff
@@ -65,3 +65,35 @@ type Extrinsic @entity {
...
+
+type TokenDocument @jsonField {
+  name: String!
+  link: String!
+}
+
+type Token @entity {
+  id: ID! # ticker
+  ticker: String! @index(unique: true)
+  name: String
```
polymath-eric, line 82: I think name should get an `@index(unique: false)`

src/mappings/entities/mapToken.ts
```diff
@@ -0,0 +1,165 @@
+import BN from 'bn.js';
+import { SubstrateExtrinsic } from '@subql/types';
+import { ModuleIdEnum, CallIdEnum } from './common';
+import { Token } from '../../types';
+
+const chainNumberMultiplier = new BN(1000000);
+
+export async function mapToken(
```
polymath-eric, line 8: FYI Sonar cloud is saying this function is too big. To make it happy each handler could be its own function and this just routes the event.

src/mappings/entities/mapToken.ts
```diff
@@ -0,0 +1,165 @@
...
+      holders: [],
+      totalSupply: '0',
+    }).save();
+  }
+
+  if (callId === CallIdEnum.RenameAsset) {
+    const { ticker, name: newName } = params;
+    const token = await Token.getByTicker(ticker);
+    if (!token) return;
+    await Token.create({
```
polymath-eric, line 52: What about, 
```
token.name = newName
await token.save()
```
?

I don't think we should  reinstantiate an instance we already have.

(applies to each case)

src/mappings/entities/mapToken.ts
```diff
@@ -0,0 +1,165 @@
...
+      ...token,
+      fundingRound: newFundingRound,
+    }).save();
+  }
+
+  if (callId === CallIdEnum.AddDocuments) {
+    const { ticker, docs } = params;
+    const token = await Token.getByTicker(ticker);
+    if (!token) return;
+    const formattedDocs = docs.map((d: any) => ({
```
polymath-eric, line 72: Does AddDocuments actually replace existing ones?

src/mappings/entities/mapToken.ts
```diff
@@ -0,0 +1,165 @@
...
+      fullAgents: [ownerDid],
+      holders: [],
+      totalSupply: '0',
+    }).save();
+  }
+
+  if (callId === CallIdEnum.RenameAsset) {
+    const { ticker, name: newName } = params;
+    const token = await Token.getByTicker(ticker);
+    if (!token) return;
```
polymath-eric, line 51: If this happens then the DB is in a bad state. We should throw an error in this case.

docker-compose.yml
```diff
@@ -14,21 +14,20 @@ services:
...
       NODE_ENV: local
       DB_HOST: postgres
-      #START_BLOCK: 3566427
       DB_PORT: 5432
-      NETWORK_ENDPOINT: ws://host.docker.internal:9944
-      #NETWORK_ENDPOINT: wss://itn-rpc.polymesh.live
-      #NETWORK_ENDPOINT: wss://alcyone-rpc.polymesh.live
+      START_BLOCK: 1362830
+      NETWORK_ENDPOINT: wss://testnet-rpc.polymesh.live
+      # NETWORK_ENDPOINT: ws://host.docker.internal:9944
```
polymath-eric, line 30: I think this should be kept as the docker host and start at block 1

schema.graphql
```diff
@@ -65,3 +65,35 @@ type Extrinsic @entity {
...
+  did: String!
+  amount: String!
+}
+
+type TokenDocument @jsonField {
+  name: String!
+  link: String!
+}
+
+type Token @entity {
```
polymath-eric, line 79: In the SDK we are trying to use the term `Asset` instead of `Token`.

Would you mind using that language here as well, i.e. `Token` -> `Asset`?

src/mappings/util.ts
```diff
@@ -65,8 +75,24 @@ export const findTopLevelCommas = (
...
   return commas;
 };
+
+export const formatParams = (
+  params: HarvesterLikeCallArgs,
+  formatKey = true,
+): Record<string, any> => {
+  const obj: Record<string, any> = {};
+  params.forEach((p) => {
+    obj[formatKey ? snakeToCamelCase(p.name) : p.name] =
```
polymath-eric, line 90: was this just copied from the other branch?

honarpour: Most of them yes but this format util was brought over from ST Viewer repo, we used it to format GQL results to key/value object. Here I'm using it to do the same to the params and send it to map function.

honarpour: I copied stuff from subquery branch to make merging later easier. We can ignore most stuff here and mainly just take the new functions and plug them in worst case.

schema.graphql
honarpour: Yeah that better, will update.

src/mappings/entities/mapToken.ts
honarpour: True. I added it as fail safe, wasn't sure if it'll break the whole thing if it happens. We could implement error reporting behind the scenes for some thing like this. Or is it fine to throw an error?

src/mappings/entities/mapToken.ts
honarpour: Yes, every time a user makes a change to documents from UI, all existing are removed first, then all new are submitted to add extrinsic. This creates an edge case when they remove all docs, or just remove one via Polkadot, in this case there's no add call coming afterwards so docs don't get updated. We need to find and store doc ID when adding, so that we can also handle remove doc call (doc IDs are the params for remove).

src/mappings/entities/mapToken.ts
honarpour: Ah cool, didn't know I can do that. Didn't see it docs either. But it looked liked every time you do `.create().save()` it basically upserts.

src/mappings/entities/mapToken.ts
honarpour: There's a bunch more stuff left 😁 
Cool, will refactor.

schema.graphql
```diff
@@ -65,3 +65,35 @@ type Extrinsic @entity {
...
+  id: ID! # ticker
+  ticker: String! @index(unique: true)
+  name: String
+  type: String
+  fundingRound: String
+  isDivisible: Boolean!
+  isFrozen: Boolean!
+  isUniquenessRequired: Boolean!
+  documents: [TokenDocument]!
+  identifiers: String! # JsonString
```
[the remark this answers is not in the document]
honarpour: Yeah this is temporary. Will define like `TokenDocument`.

src/mappings/entities/mapToken.ts
polymath-eric: I think its fine to throw an error. It will stop indexing / break it, but I think its better than showing wrong or missing data.

Considering how this gets deployed, we should still have the old version up

I am pretty sure we throw errors in the native branch so it should be alright.

src/mappings/util.ts
honarpour: Would also need `CallIdEnum` from `mappings/entities/common.ts`.

schema.graphql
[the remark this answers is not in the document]
honarpour: Updated.

src/mappings/entities/mapToken.ts
honarpour: Found document id in add document call events. It now also handles remove document call.

honarpour:
Closing this in favour of #44 going into `develop`.

PR #44, opened 2022-01-27
feat: Add asset entity

honarpour:
This PR:
- [x] Add call ID enums: `CallIdEnum`
- [x] Add asset entity to GQL schema: `Asset`
- [x] Add asset map util to extrinsic handler to update basic token info: `handleCall`, `mapAsset`
- [x] Handle updating asset holders and total transfers (settlements)
- [x] Handle asset ownership transfers
- [x] Handle compliance rules
- [x] Handle transfer managers


src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,521 @@
...
+        from: l.from.did.toString(),
+        to: l.to.did.toString(),
+        ticker: l.asset.toString(),
+        amount: chainAmountToBigNumber(l.amount).toString(),
+      }).save();
+    }),
+  );
+};
+
+const handleAffirmInstruction = async (params: Record<string, any>) => {
```
polymath-eric, line 220: Each participant will affirm an instruction.

I think we want to wait for a completed instruction and not the affirmation for this.

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,521 @@
...
+        {
+          did: leg.to,
+          amount: currentToAmount.plus(settlementAmount).toString(),
+        },
+      ];
+      asset.totalTransfers = new BigNumber(asset.totalTransfers)
+        .plus(new BigNumber(1))
+        .toString();
+      await asset.save();
+      await Settlement.remove(leg.id);
```
polymath-eric, line 247: Can you explain why we'd want to remove legs as they are being completed? I am thinking we will still want to query them.

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,521 @@
...
+        },
+        {
+          did: leg.to,
+          amount: currentToAmount.plus(settlementAmount).toString(),
+        },
+      ];
+      asset.totalTransfers = new BigNumber(asset.totalTransfers)
+        .plus(new BigNumber(1))
+        .toString();
+      await asset.save();
```
polymath-eric, line 246: I think the asset.save + Settlement.remove can be done in parallel with a `Promise.all()`

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,521 @@
...
+  await Authorization.remove(authId);
+};
+// #endregion
+
+// #region ModuleIdEnum.Externalagents
+const handleAcceptBecomeAgent = async (params: Record<string, any>) => {
+  const { authId } = params;
+  const authorization = await getAuthorization(authId);
+  const authorizationData = JSON.parse(authorization.data || '{}');
+  if (!authorization.ticker || authorizationData.group !== 'Full') {
```
polymath-eric, line 310: I think `authorization` or `authorization.data` should have a `type` field to check instead of duck typing on the fields.

If not, can you comment what authorizations you are checking for.

polymath-eric (commented):
Overall looks good.

I think the biggest thing I notice is around affirmations. An affirmation means one participant approves the transaction, but nothing happens until all parties affirm the settlement.

src/mappings/entities/mapAsset.ts
honarpour: In this case it's affirming an instruction that was affirmed when added (`addAndAffirmInstruction`). This works with the UI but yeah would be better to also handle cases where someone might do `addInstruction` from Polkadot and affirm it later.

src/mappings/entities/mapAsset.ts
honarpour: I was thinking we only care about "active" settlements, we also have tx history. But I can add a boolean field to indicate it's been processed vs remove.

src/mappings/entities/mapAsset.ts
polymath-eric: even with `addAndAffirmInstruction`, the other party still needs to affirm it, so unless its an instruction thats acting a portfolio transfer wouldn't we run into the multi affirm issue?

src/mappings/entities/mapAsset.ts
honarpour: I wanted to make sure the asset saves successfully before removing the settlement and losing it. Which shouldn't be an issue if we keep settlements.

src/mappings/entities/mapAsset.ts
honarpour: `authorization` does have a `type` field. In the `handleAcceptBecomeAgent` function, and for now, we only care about authorization to become full agent, so we're checking for that. Since each authorization type has a different set of params, they're being stored as a JSON string in the `data` field.

src/mappings/entities/mapAsset.ts
polymath-eric: I think everything is wrapped in a SQL transaction per block by the framework so it _should_ be an all or nothing as it is.

I am not 100% sure thats the case though.

src/mappings/entities/mapAsset.ts
honarpour: Yeah and this function (`handleAffirmInstruction`) handles the other party's affirmation. The initial party already affirmed it when they added it. We're only listening to `addAndAffirmInstruction` right now, and handling it using a separate function (`handleAddAndAffirmInstruction`). So this will handle all settlements that were added via `addAndAffirmInstruction`, which is what the UI does.

src/mappings/entities/mapAsset.ts
polymath-eric: It might be OK if we have a different way to get at the data, but I think if I was querying the Settlement it would be weird to see legs vanish.

src/mappings/entities/mapAsset.ts
polymath-eric: What about instructions with multiple parties involved? There can be arbitrary parties involved in an instruction, I don't think we can assume its always 2 people. There can be a complex one with 5 parties involved and we'd get 4 affirmation in addition to the AddAndAffirm.

honarpour: Each party would be a different leg, right? How does it work on the chain side? Because you get one instruction ID for the legs. So when you affirm an instruction by providing the ID, does the chain wait for all parties to affirm the ID?
It handles multiple legs, but as soon as someone affirms it, it will process it.

honarpour: Apparently the chain will process each leg as it is affirmed. So will implement the change to process the leg for user who called affirm.

polymath-eric: There is one ID for everything, legs don't get an ID. My understanding is when you affirm an instruction the chain only marks that and only triggers the settlement if that was the last pending affirmation.

I looks like there is an  `InstructionExecuted` event, that we can use. When that fires all the legs have been executed atomically.

I think we also need `InstructionFailed`, `InstructionRejected` + `InstructionRescheduled` (Not sure if this would trigger a Created event as well though) events to keep track of Settlements that are complete as well.

polymath-eric: An affirmation causes is a very minimal state change unless it was the last pending one.

Then all the legs get executed and the assets get moved only on the Executed event.

polymath-eric: The point of having legs is to be able to facilitate complex transactions atomically. If they were executed one by one there wouldn't be much use for them imo.

src/mappings/entities/mapAsset.ts
honarpour: Refactored to use existing settlements on event side.

src/mappings/entities/mapAsset.ts
honarpour: Refactored to use existing settlements on event side.

schema.graphql
```diff
@@ -296,3 +296,65 @@ type IssuerIdentityWithClaims @entity {
...
+  name: String!
+  link: String!
+}
+
+type AssetIdentifier @jsonField {
+  type: String!
+  value: String!
+}
+
+type AssetPendingOwnership @entity {
```
polymath-eric, line 316: (nit) 

I feel like this would be clearer with the word Transfer even if its a bit long. I think using Auth instead of pending works to, since its a core concepts. e.g.`AssetOwnershipTransferAuth`

polymath-eric (changes requested):

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,453 @@
...
+  const asset = await getAsset(ticker);
+  asset.isDivisible = true;
+  await asset.save();
+};
+
+const handleIssue = async (params: Record<string, any>) => {
+  const { ticker, amount } = params;
+  const asset = await getAsset(ticker);
+  const formattedAmount = chainAmountToBigNumber(amount);
+  const ownerAmount = getHolderAmount(asset.ownerDid, asset.holders);
```
polymath-eric, line 137: I don't think the owner is always the one who will receive the issuance: https://polymathnetwork.slack.com/archives/C01RLA867C5/p1644335531529869?thread_ts=1644331664.658489&cid=C01RLA867C5

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,453 @@
...
+    from,
+    to,
+    type,
+    data,
+  }).save();
+};
+
+const handleRemovePendingOwnership = async (params: Record<string, any>) => {
+  const { authId } = params;
+  await AssetPendingOwnership.remove(authId).catch(() => {
```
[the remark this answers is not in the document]
honarpour: Good point

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,453 @@
...
+};
+
+export async function mapAsset(
+  blockId: number,
+  callId: CallIdEnum,
+  moduleId: ModuleIdEnum,
+  params: Record<string, any>,
+  extrinsic: SubstrateExtrinsic
+): Promise<void> {
+  if (!extrinsic.success) return;
```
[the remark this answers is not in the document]
honarpour: I'm not 100% sure whether extrinsic handler would fire on fails or not. With events, we're checking for specific outcomes. For example, `InstructionExecuted` which means the tx ran and it worked. So to be sure, I added that check. Can look into it more to confirm.

schema.graphql
```diff
@@ -296,3 +296,65 @@ type IssuerIdentityWithClaims @entity {
   did: String! @index
   claims: [ClaimJson!]!
 }
+
+type AssetHolder @jsonField {
```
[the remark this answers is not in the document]
honarpour: Yeah I added it to settlements to keep track of the asset's holders, that part makes sense, but I agree with the performance concern.

schema.graphql
honarpour: I initially called it auths but since its use case is very narrow now, I called it what it's being used for. The thing is when transferring asset ownership, 2 types of auths are created, ownership transfer and become agent. I also added the word pending to indicate it's a temporary state.
How about `AssetPendingOwnershipTransfer`?

polymath-eric: I like `AssetPendingOwnershipTransfer`

src/mappings/entities/mapAsset.ts
honarpour: You're right, technically it's the PIA, which is now external/full agent, no?
I'll work on it but this should be fine for now since owner/PIA are currently the same, and we're transferring full agency with ownership.

schema.graphql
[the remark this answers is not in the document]
polymath-eric: I mean it makes sense to put the code in settlements, but that also suggest this is important enough to be its own Entity.

src/mappings/entities/mapAsset.ts
polymath-eric: If the constraint is that the PIA is always the owner then this is OK, but I thought since we have the term PIA that implies they might not always be the same.

schema.graphql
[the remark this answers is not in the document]
honarpour: Fair enough.

src/mappings/entities/mapAsset.ts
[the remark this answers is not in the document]
honarpour: Turns out it'll just ignore it if not exists, no error is triggered, so removed the catch.

src/mappings/entities/mapAsset.ts
honarpour: Let's revisit this later. With the new projects kicking off we'll need this new asset entity.

polymath-eric (approved):
LGTM

polymath-eric:
Looks like the test suite is getting stuck on a block. Let me know if you want help debugging it.

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,436 @@
...
+  const asset = await getAsset(ticker);
+  asset.isDivisible = true;
+  await asset.save();
+};
+
+const handleIssue = async (params: Record<string, any>) => {
+  const { ticker, amount } = params;
+  const asset = await getAsset(ticker);
+  const formattedAmount = chainAmountToBigNumber(amount);
+  const assetOwner = await AssetHolder.get(`${asset.ownerDid}:${ticker}`);
```
polymath-eric, line 136: This isn't guaranteed to be found.

I think this needs to be wrapped with a `findOrCreate` type method.

src/mappings/entities/mapSettlement.ts
```diff
@@ -183,6 +185,23 @@ async function handleInstructionFinalizedEvent(
...
   });
   await Promise.all([settlement.save(), instruction.save()]);
+
+  if (eventId === EventIdEnum.InstructionExecuted) {
+    await Promise.all(
+      instruction.legs.map(async ({ ticker, amount, from, to }) => {
+        const asset = await getAsset(ticker);
+        const settlementAmount = new BigNumber(amount);
+        const fromHolder = await AssetHolder.get(`${from.did}:${ticker}`);
+        const toHolder = await AssetHolder.get(`${to.did}:${ticker}`);
```
polymath-eric, line 195: These aren't guaranteed to be found. I am thinking we need a findOrCreate wrapper.

As is the tests fail here because of this.

src/mappings/entities/mapAsset.ts
```diff
@@ -0,0 +1,436 @@
...
+};
+// #endregion
+
+// #region ModuleIdEnum.Identity
+const handleAddPendingOwnership = async (params: Record<string, any>, extrinsic: any) => {
+  const { target: targetData, authorizationData } = params;
+  const id = extrinsic.events[0].event.data[3].toString();
+  let ticker: string;
+  const from = extrinsic.events[0].event.data[0].toString();
+  const to = targetData.Identity.toString();
```
polymath-eric, line 183: An authorization can target an Identity or an Account, so this should be something like:

```
let to: string;
  if (!targetData.Identity.isEmpty) {
    to = targetData.Identity.toString();
  } else {
    to = targetData.Account.toString();
  }
  ```
  Preferably wrapped in a helper. It might be the case the Asset transfer only targets Identities, regardless I think it's good to be defensive and be able to handle both cases here since thats what the type permits.

File: src/mappings/entities/mapAsset.ts [hunk 13]
```diff
@@ -0,0 +1,436 @@
...
+  const type = Object.keys(authorizationData)[0];
```
polymath-eric, line 184: Move this type extraction near the top of the function and make sure it equals `TransferAssetOwnership` or `BecomeAgent`. 

Block 102 was failing because it was trying to process a `JoinIdentity` auth that was targeting an account.

Also they should be enum like values, I don't think we have them yet, but we should probably add Authorization types into `src/mappings/entities/common.ts`

polymath-eric (changes requested):
In addition to the comments please add integration tests in `tests/integration`. Look at the existing tests for examples.

src/mappings/entities/mapAsset.ts
honarpour: Ownership transfer and full agency is only done using the DID. Since we're now checking for the auth types specific to ownership transfers, this shouldn't be an issue. We're also only keeping track of DIDs right now. Handling accounts would add complexity since multiple accounts could be tied to the same DID. We can add it later if required.

polymath-eric: Makes sense, I was thinking we'd just use the address instead, but that doesn't make sense for the asset holder.

I guess the main thing was to check for the right types of auths.

polymath-eric (approved):
LGTM

PR #45, opened 2022-01-31
Subquery native graphql (resolved conflicts)

honarpour:
We can just merge this instead of #30, I merged `develop` into `subquery_native_graphql` locally and resolved conflicts. The branch was protected so had to push to a new branch.


polymath-eric (approved):
LGTM

For transparency, I looked at the diff against what was in subquery_native branch

I didn't read all of the code here thoroughly but I did skim it. The PRs were all reviewed as they've been merged into that branch. Let me know if there is any part that warrants extra attention.

polymath-eric:
SonarCloud has three smells which I think are all reasonable, they should be easy to fix.

PR #46, opened 2022-02-07
feat: 🎸 Update to support 4.1.1 chain version

polymath-eric:
The previous 3002 spec version is now 3010, this is to give the 4.0.x
chain versions space to upgrade

### DIFF
--- a/project.template.yaml
+++ b/project.template.yaml
@@ -80,7 +80,7 @@ network:
       'AuthorityList': 'Vec<NextAuthority>',
       'ElectionCompute':
         { '_enum': ['OnChain', 'Signed', 'Authority'] }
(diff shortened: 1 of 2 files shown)


prashantasdeveloper (approved):

project.template.yaml
```diff
@@ -1049,6 +1049,7 @@ network:
                   'ItnRewardStatus': { '_enum': { 'Unclaimed': 'Balance', 'Claimed': '' } },
                 },
             },
+            { minmax: [3002, 3002], types: *4.1.x },
```
VictorVicente, line 1052: Are you sure this line would work? Have you used it before?

VictorVicente (approved):
Once this one get merged, you should create the required tag

PR #50, opened 2022-02-22
feat: 🎸 Add support for spec 3010

polymath-eric:
Bring the update spec version stuff into develop so future work has it.


polymath-eric:
Already had these changes in develop

PR #53, opened 2022-03-07
fix: 🐛 Fix mishandled InstructionFinalized event

polymath-eric:


src/mappings/entities/mapSettlement.ts
```diff
@@ -140,7 +139,7 @@ async function handleInstructionUpdate(params: Codec[], event: SubstrateEvent) {
   const instructionId = getTextValue(params[2]);
   const instruction = await Instruction.get(instructionId);
   if (instruction) {
-    instruction.addresses.push(signer);
+    if (signer) instruction.addresses.push(signer);
```
polymath-eric, line 142: This is being done in the handlenstructionFinalized method, so I figure I'd put it here just to be safe.

src/mappings/entities/mapSettlement.ts
```diff
@@ -24,7 +24,6 @@ const updateEvents: EventIdEnum[] = [
   EventIdEnum.InstructionAuthorized,
   EventIdEnum.InstructionUnauthorized,
   EventIdEnum.InstructionAffirmed,
-  EventIdEnum.InstructionFailed,
```
polymath-eric, line 27: InstructionFailed isn't an extrinsic, but something the gets triggered and doesn't have a signer.

To prevent the block, since handleInstructionUpdate is only adding the signers, its save to0 exclude from there. Its being handled in the finalizedEvents

honarpour (approved):

PR #54, opened 2022-03-21
feat: 🎸 Initial schema redesign

prashantasdeveloper:
### Description

* Refactor usage of `blockId` with `block` attribute in all entities
* Add `unique: false` to all indexes
* Add identities related entities (`Identity`, `Account`, `Permissions`)
* Remove grouping related claim entities (`IssuerIdentityWithClaims` and `IdentityWithClaims`)
* Add `BridgeEvent` entity
* Add migration file for setup of test environments 

### JIRA Link
DA-240, DA-309

### Checklist


schema.graphql
```diff
@@ -254,47 +255,53 @@ type Scope @jsonField {
...
-  scope: Scope
-  cddId: String
-}
-
-type ClaimJson @jsonField {
-  targetDid: String!
-  issuer: String!
+  block: Block! @index(unique: false)
+  eventIdx: Int! @index(unique: false)
+  targetDid: IdentityWithClaims! @index(unique: false)
```
polymath-eric, line 267:
```suggestion
  target: IdentityWithClaims! @index(unique: false)
```
This is given us `targetDidId`, which is a weird doubling of "Id". DID is the ID and now its a reference to an Entity.

src/mappings/entities/mapClaim.ts
```diff
@@ -18,7 +15,7 @@ enum ClaimScopeTypeEnum {
 }
 
 type ClaimParams = {
-  claimExpiry: bigint;
+  claimExpiry: bigint | undefined;
```
polymath-eric, line 18: I think this means the same thing, but if not OK to leave.
```suggestion
  claimExpiry?: bigint;
```

polymath-eric (changes requested):
Requesting changes for `targetDid` should become just `target` imo.

I'd prefer to drop parentBlock altogether. If we know the clients, we should be able to add in an alias (GraphQL feature) on the query and not need to dig into their code. We already will have to update their config for the switch. Let me know if there is another reason I am not considering for this point though.

Other comments are nits.

schema.graphql
```diff
@@ -254,47 +266,53 @@ type Scope @jsonField {
 
 type ClaimScope @entity {
   id: ID!
-  targetDid: String! @index
+  target: String! @index(unique: false)
```
polymath-eric, line 269: This one could have been left as targetDid, since its not a reference type.

OK to leave if you think `target` is better though.

polymath-eric (commented):
Overall LGTM.

Seems like snapshot needs to be updated + an unused import removed.

tests/integration/__snapshots__/authorizations.test.ts.snap
```diff
@@ -196,86 +27,7 @@ exports[`authorizations should return authorisations filtered by authorization t
...
-        "expiry": null,
-        "fromDid": "0x0500000000000000000000000000000000000000000000000000000000000000",
-        "status": "Pending",
-        "toDid": "0xee7cd3d29832c3ad6db1dedb99b8fd146b57c6458fd79d17c5ac81f3d098227a",
-        "toKey": null,
-        "type": "AddMultiSigSigner",
-        "updatedBlock": 104,
-      },
-    ],
+    "nodes": Array [],
```
polymath-eric, line 30: This shouldn't be an empty array, it probably means something is wrong here.

polymath-eric (commented):
The snapshots that are empty arrays now don't look right to me. For the tests to be meaningful they need to return some data.

There is probably a problem thats causing blocks to not be processed.

tests/integration/__snapshots__/authorizations.test.ts.snap
prashantasdeveloper: Can you once run the tests in your machine once? CI also has passed these so I'm not sure what to do here. Maybe the snapshot needs to be updated or something in `polymesh-local`

polymath-eric: I'll give it a shot.

With snapshot based tests passing a `-u` will make them almost always pass a CI run, the value comes in looking at the diff.

scripts/.env
```diff
@@ -10,6 +10,6 @@ MYSQL_USER=root
 MYSQL_PASSWORD=root
 MYSQL_DATABASE=itn
 
-START_BLOCK=0
+START_BLOCK=2355800
```
polymath-eric: What is reading this value? I wonder if this is interfering with the tests somehow.

Pushed a change to make this 0 again to see if that changes anything.

polymath-eric (changes requested):
I got this error on testnet, so we need to look into this more before merging:

```
ERROR failed to index block at height 529173 handleEvent() SequelizeDatabaseError: index row size 2752 exceeds btree version 4 maximum 2704 for index "issuer_identity_with_claims_target_index"
```

I can look into this more, but have staging to look at first.

schema.graphql
```diff
@@ -1,15 +1,15 @@
...
   eventIdx: Int!
   extrinsicIdx: Int
   specVersionId: Int!
   moduleId: String! @index(unique: false)
   eventId: String! @index(unique: false)
   attributesTxt: String!
-  eventArg_0: String
-  eventArg_1: String
-  eventArg_2: String
+  eventArg_0: String @index(unique: false)
```
polymath-eric, line 10: Remove. Can't index event_args with a btree which this generates

polymath-eric: I think this comment still stands, although I don't quite remember my reasoning from last month. 

Some events like batch can have very large inputs, but BTree indexes have a limited size.

If we want to index something like this I think it needs to be a GIN.

VictorVicente (approved):

src/mappings/entities/mapClaim.ts
```diff
@@ -1,120 +1,152 @@
...
+    scope,
+    jurisdiction,
+    cddId,
+    filterExpiry,
+  }).save();
+
+  if (scope) {
+    await handleScopes(
+      target,
+      scope.type === ClaimScopeTypeEnum.Ticker ? scope.value : null,
```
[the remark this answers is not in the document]
prashantasdeveloper: `handleScopes(target, ticker, scope?)` is called from two places that is why the ticker part is calculated here.

PR #56, opened 2022-03-25
fix: 🐛 Revert to using snake case in extrinsics table

polymath-eric:
Tooling GQL expects snake case for call_id

Reverts the values to what they were [before](https://github.com/PolymathNetwork/polymesh-subquery/compare/v4.1.0-s1...v5.0.0#diff-cacfd1ec0155919c9a4ad17775cce37b7d92a736f89d84216e087d8f19405385L132)

Note the column names remain snake case in the DB despite the entity field being camel case.


polymath-eric:
Merging this in without review as its for a production fix and relatively small.

PR #57, opened 2022-03-31
feat: Revamp Asset related entities

polymath-eric:
### Description
* Map asset related entities with events instead of extrinsics. (This removes the usage of `handleToolingCall` and directly `handleCall` method is invoked from the manifest file for call handler)
* Revamp asset related entities to convert json fields to their respective entities. Following new entities have been created:
  * AssetDocument
  * Compliance
  * TransferManager
* Also, added tests for all asset related entities

### JIRA Links
DA-267, DA-313

### Checklist


src/mappings/util.ts
```diff
@@ -138,28 +149,92 @@ export const getSigner = (extrinsic: SubstrateExtrinsic): string => {
...
+  };
+};
+
+/**
+ * Parses AssetTransferManager
+ */
+export const getTransferManagerValue = (
+  manager: Codec
+): Pick<TransferManager, 'type' | 'value'> => {
+  const { countTransferManager, percentageTransferManager } = JSON.parse(JSON.stringify(manager));
```
polymath-eric, line 216: 5.0.0 renamed these to be "Conditions" instead of "Manager". Try to test these against 5.0.0

We might need to support both names for 4.0.0 and 5.0.0 chains.

Long term I think just the new name will be OK since the storage should get migrated on chain.

src/mappings/util.ts
```diff
@@ -138,28 +149,92 @@ export const getSigner = (extrinsic: SubstrateExtrinsic): string => {
...
+    return {
+      type: TransferRestrictionType.Count,
+      value: Number(countTransferManager),
+    };
+  }
+
+  if (percentageTransferManager) {
+    return {
+      type: TransferRestrictionType.Percentage,
+      value: Number(countTransferManager),
```
polymath-eric, line 228: `percentageTransferManager` ?

polymath-eric (commented):
Two things around transfer managers, but looks good otherwise.

src/mappings/util.ts
prashantasdeveloper: This is a sample of params received. 
for % -
`[{"value":"0x7b5de4a8dfb706a24f5288082c3e4efdd834309785f7bee95ecea0d8f3abd80a"},{"value":"UNDVIDED"},{"value":{"PercentageTransferManager":50000}}]`

for count - `[{"value":"0x2e03244bbd44db4320b0a992f7e8eda9c88931be6c2e828b86147228531cf280"},{"value":"UTP-TC-0001"},{"value":{"CountTransferManager":45000}}]`

src/mappings/entities/mapTransferManager.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+};
+
+export async function mapTransferManager(
+  eventId: EventIdEnum,
+  moduleId: ModuleIdEnum,
+  params: Codec[]
+): Promise<void> {
+  if (moduleId === ModuleIdEnum.Statistics) {
+    if (eventId === EventIdEnum.TransferManagerAdded) {
```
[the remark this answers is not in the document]
polymath-eric: Makes sense

src/mappings/util.ts
polymath-eric: wouldn't countTransferManager be undefined here then? (also surprised at the PascalCase vs camelCase being OK)

polymath-eric (commented):
LGTM

(I can't approve since I am the author)

prashantasdeveloper (approved):

PR #58, opened 2022-04-01
feat: 🎸 Add tooling only mode

polymath-eric:
Allow for TOOLING_ONLY env to be set that will skip native GraphQL
handlers. This should allow for more stable indexing for tooling-gql


src/mappings/mappingHandlers.ts
```diff
@@ -138,32 +199,28 @@ export async function handleEvent(event: SubstrateEvent): Promise<void> {
     })
   );
 
-  await Event.create({
-    id: `${blockId}/${eventIdx}`,
+  await Promise.all(handlerPromises);
+}
+
+// handles calls to populate native GraphQL tables as well as what is needed for tooling
```
polymath-eric, line 205: When I get the asset handler migrated, this can be reverted to just one handler, so only events is weird

prashantasdeveloper (commented):
Overall LGTM, but I think we should add indexes in other columns-
- Event (blockId, eventArg_0, eventArg_1, eventArg_2), 
- Block (blockId)
- Extrinsic (blockId, moduleId, callId, success, extrinsicHash)

prashantasdeveloper:
Also, while testing tooling-gql with testnet, I received the following error while fetching events - `Cannot return null for non-nullable field Scope.type`. Some of the rows in events had claim_scope as `{"type":null,"value":null}`.

Is this even allowed? We should probably confirm this during the standup and make changes in the function that handles claim scopes accordingly

polymath-eric:
Good catch with the `null` claims. It was like that in develop branch.

I updated it and seems like it fixes tooling.

polymath-eric:
@prashantasdeveloper Do we need to have indexes on those columns? Seems like tooling needs composite indexes which get added in with `compat.sql`

I am OK adding them if we need them, but if not, I'd like to revert that commit. Having it on event and block is going to trigger on every block and increase our sync times.

prashantasdeveloper (approved):
LGTM.

For now the indexes added in `schema.graphql` would be enough. We don't need any more indexes.(Mostly the missing index on `block_id` do the trick here). We could probably monitor once this change goes live and still needs be that more indexes are required, we can add them later.

VictorVicente (approved):

PR #61, opened 2022-04-18
feat: 🎸 5.0.0 support

prashantasdeveloper:
### Description

- Added support for types spec version `5_000_000`
- Migrated subquery-node version to 1.0.0
- Migrated `project.yaml` to spec version `1.0.0`
- Removed unwanted dependencies from package.json

### JIRA Link

DA-262

### Checklist


project.template.yaml
```diff
@@ -96,6 +96,1019 @@ network:
                 },
             },
             ### </Harvester compatibility special cases>
+            {
+              minmax: [5_000_000, 5_000_000],
```
polymath-eric, line 100: With a number this large I'd think there be a range where new types won't be updated e.g. `[5_000_000, 5_000_999]`. Not sure if thats the actual range, but can you check and bump up the max if thats the case?

It would be be nice to be able to not need to redeploy if we don't have to.

VictorVicente: Yes, we have such a discussion. I'll provide you the document Core team prepared to talk about this.

Dockerfile
```diff
@@ -8,10 +8,14 @@ COPY package.json /app
...
 RUN yarn --frozen-lockfile && chown -R subquery /app
 
+########################################################################################
+# Following parameters are needed to generate a project.yaml from project.template.yaml
 ENV START_BLOCK=1
-# CALL_HANDLER and EVENT_HANDLER are needed to generate a project.yaml from project.template.yaml
 ENV CALL_HANDLER='callHandler'
 ENV EVENT_HANDLER='handleEvent'
+# ENV NETWORK_ENDPOINT='ws://host.docker.internal:9944'
+# ENV NETWORK_CHAIN_ID='0xda7f2072787bfd0b09f7e12fca619afb6041b3d620f39f3a508814869100bf01'
```
polymath-eric, line 17: should these be uncommented?

Could be nice to have a line about how to get chain id, or if it defaults to the right thing.

src/mappings/entities/mapInvestment.ts
```diff
@@ -20,11 +20,11 @@ export async function mapInvestment(
       id: `${blockId}/${event.idx}`,
       blockId,
       investor: getTextValue(params[0]),
-      stoId: params[1],
+      stoId: Number(params[1].toString()),
       offeringToken: serializeTicker(params[2]),
       raiseToken: serializeTicker(params[3]),
-      offeringTokenAmount: getTextValue(params[4]),
-      raiseTokenAmount: getTextValue(params[5]),
+      offeringTokenAmount: BigInt(getTextValue(params[4])),
```
polymath-eric, line 26: What about having a `getBigIntValue` helper?

src/mappings/entities/mapProposal.ts
```diff
@@ -53,8 +53,8 @@ export async function mapProposal(
   if (eventId === EventIdEnum.Voted) {
     const account = getTextValue(params[1]);
     const pipId = getTextValue(params[2]);
-    const vote = getTextValue(params[3]);
-    const weight = getTextValue(params[4]);
+    const vote = JSON.parse(getTextValue(params[3])) as boolean;
```
polymath-eric, line 56: `getBooleanValue` ?

polymath-eric (commented):
LGTM, left a few comments about having a few more getValue helper functions.

Dockerfile
prashantasdeveloper: Yes, I completely forgot about this. So the point was I was trying to re-create the network section within the project.yaml file as we create the handlers(using $CALL_HANDLER and $EVENT_HANDLER).So I introduced these vars so that would just pass them in the docker file, but apparently when I try to do this, the validations within the subquery fails stating chainId not found, even though the project.yaml file is perfectly created using the envsubst.

Can you try it out once Eric? I'm not that good with docker. Maybe I missed something. I was hoping to achieve a network section as follows - 
```
network: 
  chainId: ${NETWORK_CHAIN_ID}
  endpoint: ${NETWORK_ENDPOINT}
```

prashantasdeveloper: https://github.com/subquery/subql/blob/main/packages/common-substrate/src/project/versioned/v1_0_0/model.ts#L64 This is where it fails stating something like - cannot find value.trim()

package.json
```diff
@@ -27,24 +27,18 @@
...
-    "@polkadot/api": "5.1.1",
-    "@polkadot/util": "7.2.1",
-    "@polkadot/util-crypto": "7.2.1",
+    "@polkadot/api": "^8",
     "bignumber.js": "^9.0.2",
-    "bn.js": "^4.11.9",
     "typeorm": "^0.2.34"
   },
-  "author": "raycar5",
+  "author": "Prashant",
```
polymath-eric, line 34: Should this be "Polymath" or something along those lines?

tsconfig.json
```diff
@@ -16,10 +16,13 @@
       // this is the package name we use (in the interface imports, --package for generators) */
       "polymesh-subql/*": ["src/*"],
       // here we replace the @polkadot/api augmentation with our own, generated from chain
-      "@polkadot/api/augment": ["src/api-interfaces/augment-api.ts"],
+      // "@polkadot/api/augment": ["src/api-interfaces/augment-api.ts"],
```
polymath-eric, line 19: Do we not need these anymore?

package.json
```diff
@@ -89,5 +83,8 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
+  },
+  "exports": {
+    "chainTypes": "./src/chainTypes/index.ts"
```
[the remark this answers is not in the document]
prashantasdeveloper: I followed as subquery suggested - https://university.subquery.network/build/manifest.html#working-example.

When I was directly specifying `src/chainTypes/index.ts` it was giving the following lint error - 
```
String does not match the pattern of "^\./".
```

I also got the same errors, but that is just because your build was not properly created. Since we have `incremental: true` specified in `tsconfig.json` the build is not properly updated. Can you remove your `tsconfig.tsbuildinfo` file and try again? That did the trick for me. 

Please let me know if this doesn't works

package.json
prashantasdeveloper: I guess this got replaced when I did `subql init`. I'll replace with Polymath or should it be Polymesh?

tsconfig.json
prashantasdeveloper: I don't think those were used earlier as well. We can remove those for now

package.json
polymath-eric: I am not too sure. In the SDK our error messages say please contact the Polymath team, although Polymesh makes more sense to me.

polymath-eric (commented):
Try out the ' marks in the template and see if that fixes the docker image / project template issue for you.

Dockerfile
prashantasdeveloper: Thanks for the input

docker-compose.yml
```diff
@@ -9,39 +9,66 @@ services:
...
+    volumes:
+      - ./:/app
     command:
-      - --batch-size=500
-      - -f=/app
-      - --local
+      - -f=/app # Path of the project
+      - --batch-size=100 # Batch size of blocks to fetch in one round
+      # - --scale-batch-size # scale batch size based on memory usage. To know more about the process - https://github.com/subquery/subql/blob/c9bc9733deef726f78fa48387e74e52f8d6ca8d2/packages/node/src/indexer/fetch.service.ts#L88
+      # - --timeout=3600 # Timeout for indexer sandbox to execute the mapping functions (in seconds)
```
polymath-eric, line 53: OK to leave, but an hour seems like quite a bit to me.

My preference is to have realistic values as a bit of documentation and suggestion for what config values should be. Not sure what it needs to be to deal with a large block though.

package.json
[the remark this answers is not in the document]
polymath-eric: Yeah removing tsconfig.tsbuildinfo worked for me.

We might need to put it into a docker ignore file. Its not the first time it messed me up and I forgot about the fix.

polymath-eric (approved):
LGTM

polymath-eric:
Looks like the tests aren't working. Is it because the `NO_NATIVE_GRAPHQL` flag? Looks like graphql is returning empty responses for things.

docker-compose.yml
prashantasdeveloper: I wrote that values thinking that SRE team won't have think too much if they are asked to run with any params.

As for the large block, reducing batch-size to 1 and setting timeout to 3600 worked for me. When Robert does those batch transactions we could recheck on what those block need.

polymath-eric: My concern is along the lines the connection dies (or something similar) and SQ falls behind a whole hour before reconnecting.

I think we are OK reducing the batch size down to something like 100 without impacting the sync time to much, since we need to wait for a SQL commit on every block. Some SubQuery projects can skip block entirely so probably benefit more from bigger batch sizes. I think 1 is too small and will slow down sync though.

Might be worth experimenting.

PR #63, opened 2022-05-11
feat: 🎸 Add identities related entities

prashantasdeveloper:
### Description

Add identities related entities

### JIRA Link

DA-309

### Checklist

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -66,6 +66,42 @@ type Extrinsic @entity {
   specVersionId: Int!
 }
 
+type PermissionJson @jsonField {
+  type: String!
+  values: [String]!
+}
+
+type Permissions @entity {
+  id: ID!
+  assets: PermissionJson
+  portfolios: PermissionJson
+  transactions: PermissionJson
+  transactionGroups: [String]!
+  createdBlock: Block!
+  updatedBlock: Block
+  datetime: Date!
+}
+
+type Account @entity {
+  id: ID!
+  address: String! @index(unique: true)
+  identity: Identity!
+  eventId: String! @index(unique: false)
+  permissions: Permissions!
+  block: Block!
+  datetime: Date!
+}
+
+type Identity @entity {
+  id: ID!
+  did: String! @index(unique: false)
+  primaryAccount: String! @index(unique: false)
+  secondaryAccounts: [Account!]! @derivedFrom(field: "identity")
+  secondaryKeysFrozen: Boolean!
+  block: Block!
+  datetime: Date!
+}
+
 type BridgeEvent @entity {
   id: ID!
   block: Block! @index(unique: false)
--- a/src/mappings/entities/mapIdentities.ts
+++ b/src/mappings/entities/mapIdentities.ts
@@ -1,0 +1,192 @@
+import { Codec } from '@polkadot/types/types';
+import { SubstrateEvent } from '@subql/types';
+import { EventIdEnum, ModuleIdEnum } from './common';
+import { Permissions } from '../../types/models/Permissions';
+import { Account } from '../../types/models/Account';
+import { Identity } from '../../types/models/Identity';
+import { getFirstValueFromJson, getTextValue } from '../util';
+
+/**
+ * Subscribes to the Identities related events
+ */
+export async function mapIdentities(
+  blockId: number,
+  eventId: EventIdEnum,
+  moduleId: ModuleIdEnum,
+  params: Codec[],
+  event: SubstrateEvent
+): Promise<void> {
+  const datetime = event.block.timestamp;
+  if (moduleId === ModuleIdEnum.Identity) {
+    if (eventId === EventIdEnum.DidCreated) {
+      const did = getTextValue(params[0]);
+      const address = getTextValue(params[1]);
+      await Identity.create({
+        id: did,
+        did,
+        primaryAccount: getTextValue(params[1]),
+        secondaryKeysFrozen: false,
+        blockId,
+        datetime,
+      }).save();
+
+      await Permissions.create({
+        id: address,
+        assets: null,
+        portfolios: null,
+        transactions: null,
+        transactionGroups: [],
+        createdBlockId: blockId,
+        datetime,
+      }).save();
+      await Account.create({
+        id: address,
+        identityId: did,
+        permissionsId: address,
+        eventId,
+        address,
+        blockId,
+        datetime,
+      }).save();
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysAdded) {
+      const { id: identityId } = await fetchIdentity(params[0]);
+
+      const accountsList = params[1].toJSON() as any[];
+
+      const promises = [];
+
+      for (const account of accountsList) {
+        const address = account.signer.account;
+
+        const { assets, portfolios, transactions, transactionGroups } = getPermissions(
+          account.permissions
+        );
+
+        promises.push(
+          Permissions.create({
+            id: address,
+            assets,
+            portfolios,
+            transactions,
+            transactionGroups,
+            createdBlockId: blockId,
+            datetime,
+          }).save()
+        );
+
+        promises.push(
+          Account.create({
+            id: address,
+            address,
+            identityId,
+            permissionsId: address,
+            eventId,
+            blockId,
+            datetime,
+          }).save()
+        );
+      }
+
+      await Promise.all(promises);
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysFrozen) {
+      const identity = await fetchIdentity(params[0]);
+      identity.secondaryKeysFrozen = true;
+      await identity.save();
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysUnfrozen) {
+      const identity = await fetchIdentity(params[0]);
+      identity.secondaryKeysFrozen = false;
+      await identity.save();
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysRemoved) {
+      await fetchIdentity(params[0]);
+      const accountsRemoved = params[1].toJSON() as any[];
+      const accounts = accountsRemoved.map(account => Account.remove(account.account));
+
+      await Promise.all(accounts);
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeyPermissionsUpdated) {
+      await fetchIdentity(params[0]);
+      const address = getFirstValueFromJson(params[1])['account'];
+      const permissions = await Permissions.get(address);
+      const { assets, portfolios, transactions, transactionGroups } = getPermissions(
+        params[3].toJSON() as any
+      );
+
+      permissions.assets = assets;
+      permissions.portfolios = portfolios;
+      permissions.transactions = transactions;
+      permissions.transactionGroups = transactionGroups;
+      permissions.updatedBlockId = `${blockId}`;
+      await permissions.save();
+    }
+  }
+}
+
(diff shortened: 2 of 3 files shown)


src/mappings/entities/mapIdentities.ts
```diff
@@ -0,0 +1,192 @@
...
+      permissions.transactions = transactions;
+      permissions.transactionGroups = transactionGroups;
+      permissions.updatedBlockId = `${blockId}`;
+      await permissions.save();
+    }
+  }
+}
+
+function getPermissions(accountPermissions: Record<string, unknown>) {
+  logger.info(JSON.stringify(accountPermissions));
```
polymath-eric, line 134: remove

src/mappings/entities/mapIdentities.ts
```diff
@@ -0,0 +1,192 @@
...
+  let assets,
+    portfolios,
+    transactions,
+    transactionGroups = [];
+
+  let type;
+  Object.keys(accountPermissions).forEach(key => {
+    logger.info(key);
```
polymath-eric, line 142: Should be removed. guesing this is left over from debugging.

If we are logging stuff thats not for debugging it should have more context.

src/mappings/entities/mapIdentities.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (eventId === EventIdEnum.SecondaryKeysUnfrozen) {
+      const identity = await fetchIdentity(params[0]);
+      identity.secondaryKeysFrozen = false;
+      await identity.save();
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysRemoved) {
+      await fetchIdentity(params[0]);
+      const accountsRemoved = params[1].toJSON() as any[];
```
polymath-eric, line 109: Is it possible to avoid any casts here? We should be able to cast to something more specific.

`any` basically opts out of TS checking, I think it should be avoided in the actual code. (alright occasionally in tests imo)

(few places this applies)

schema.graphql
```diff
@@ -66,6 +66,42 @@ type Extrinsic @entity {
   specVersionId: Int!
 }
 
+type PermissionJson @jsonField {
```
polymath-eric, line 69: `PermissionValue` ?

I don't think we need Json in the name (we might have it as convention already so ignore if so).

schema.graphql
```diff
@@ -66,6 +66,54 @@ type Extrinsic @entity {
...
+}
+
+type TransactionPermissions @jsonField {
+  type: String!
+  values: [String]!
+}
+
+type PortfolioPermissions @jsonField {
+  type: String!
+  values: [PortfolioJson]!
```
polymath-eric, line 81: No need for a change, but some thoughts.

I think Portfolios will get their own Entity eventually, and it would be nice to have the GraphQL edge, which I don't think we can get with JSON.

I think once we get a full Portofolio Entity we might want to migrate this. Any thoughts on this?

src/mappings/entities/mapIdentities.ts
```diff
@@ -0,0 +1,250 @@
...
+      await handleSecondaryKeysRemoved(params);
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeyPermissionsUpdated) {
+      await handleSecondaryKeysPermissionsUpdated(blockId, params);
+    }
+  }
+}
+
+const fetchIdentity = async (param: Codec): Promise<Identity> => {
```
polymath-eric, line 49: (nit) I think similar functions use `get` e.g. `getAsset`.

```suggestion
const getIdentity = async (param: Codec): Promise<Identity> => {
```

I'd be OK with a convention of `fetch` meaning async and `get` being a sync getter like `getTextValue` if thats what you are thinking, but my preference would be to keep everything with `get` since TS will let us know if something is async or not.

polymath-eric (approved):
LGTM left one nitty comment and one observational one.

src/mappings/entities/mapIdentities.ts
prashantasdeveloper: Makes sense to me.

schema.graphql
polymath-eric: Makes sense. I guess a benefit of having to resync every deploy is we don't have to worry about data migration problems.

I'd be more apprehensive if this was a normal DB and we'd have to migrate over the data.

db/migration.sql
```diff
@@ -0,0 +1,22 @@
...
+('0x0100000000000000000000000000000000000000000000000000000000000000', '0x0100000000000000000000000000000000000000000000000000000000000000', 'primaryAccount', 'f', 'DidCreated', -1, -1, now(), now(), now()) ON CONFLICT(id) DO NOTHING;
+
+--Add entry for 0x0400
+INSERT INTO "public"."identities" ("id", "did", "primary_account", "secondary_keys_frozen", "event_id", "created_block_id", "updated_block_id", "datetime", "created_at", "updated_at") VALUES
+('0x0400000000000000000000000000000000000000000000000000000000000000', '0x0400000000000000000000000000000000000000000000000000000000000000', 'primaryAccount', 'f', 'DidCreated', -1, -1, now(), now(), now()) ON CONFLICT(id) DO NOTHING;
+
+--Add entry for 0x0500
+INSERT INTO "public"."identities" ("id", "did", "primary_account", "secondary_keys_frozen", "event_id", "created_block_id", "updated_block_id", "datetime", "created_at", "updated_at") VALUES
+('0x0500000000000000000000000000000000000000000000000000000000000000', '0x0500000000000000000000000000000000000000000000000000000000000000', 'primaryAccount', 'f', 'DidCreated', -1, -1, now(), now(), now()) ON CONFLICT(id) DO NOTHING;
```
VictorVicente, line 22: Are these identities also present in Mainnet? Or is this script skipped in that case?

schema.graphql
```diff
@@ -7,9 +7,9 @@ type Event @entity {
   moduleId: String! @index(unique: false)
   eventId: String! @index(unique: false)
   attributesTxt: String!
-  eventArg_0: String @index(unique: false)
-  eventArg_1: String @index(unique: false)
-  eventArg_2: String @index(unique: false)
```
VictorVicente, line 12: Is it OK to remove the index?

src/mappings/entities/mapClaim.ts
```diff
@@ -89,75 +106,22 @@ export async function mapClaim(
...
+  if (eventId === EventIdEnum.ClaimRevoked) {
+    const id = getId(target, claimType, scope, jurisdiction, cddId);
+    const claim = await Claim.get(id);
+    if (!claim) {
+      throw new Error(`Claim with id ${id} was not found`);
+    }
+    claim.revokeDate = issuanceDate;
+    await claim.save();
+  }
+
```
VictorVicente, line 118: What do you think about refactoring this in the following shape:
- A `switch case` to make it clearer which `eventIds` it is covering and which it isn't.
- A `function` for each one so the code for the mapping is very easy to read. 

Thoughts @prashantasdeveloper @polymath-eric? If you are OK with this, it isn't necessary to apply it now, we could add a task to do it later.

src/mappings/entities/mapIdentities.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeysRemoved) {
+      await handleSecondaryKeysRemoved(params);
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeyPermissionsUpdated) {
+      await handleSecondaryKeysPermissionsUpdated(blockId, params);
+    }
+  }
```
VictorVicente, line 46: Something like this is what I meant in the previous comment!

schema.graphql
polymath-eric: We have to remove the index or we won't be able to process large extrinsics like a big batch.

By default they will be BTree which has a limit of 2713 bytes, so we'd get errors about that.

If the need to be indexed then we can add it into `compat.sql` and use a GIN or Gist (Not sure which one is best).

Right now we have indexes in that file for `left(event_arg_0, 100)` on these columns, which I think is a carry over from tooling.

db/migration.sql
prashantasdeveloper: This will be skipped for mainnet

src/mappings/entities/mapClaim.ts
prashantasdeveloper: Refactored using multiple ifs

db/migration.sql
polymath-eric: Maybe this could be called `localMigration.sql` or `localSeed.sql`? Might make is a bit more clear.

(Looks like we use snake, camel and kebab case for files here, I think camel case is most conventional for us though)

src/mappings/entities/mapClaim.ts
```diff
@@ -25,6 +22,11 @@ type ClaimParams = {
   jurisdiction: string;
 };
 
+type Scope = {
```
polymath-eric, line 25: (nit) I think interface is preferred, although it does almost the same thing as a type.

src/mappings/entities/mapClaim.ts
```diff
@@ -34,6 +36,53 @@ export async function mapClaim(
...
+const getId = (
+  target: string,
+  claimType: string,
+  scope: Scope,
+  jurisdiction: string,
+  cddId: string
+): string => {
+  const idAttributes = [target, claimType];
+  if (scope) {
+    // Not applicable in case of CustomerDueDiligence
```
polymath-eric, line 68: I think `InvestorUniquenessV2Claim` is unscoped as well.

polymath-eric (changes requested):
Changes look good, but requesting changes since the handlers are commented. (Sonar cloud didn't flag it)

src/mappings/mappingHandlers.ts
```diff
@@ -159,20 +145,23 @@ export async function handleEvent(event: SubstrateEvent): Promise<void> {
...
-    mapInvestment(...handlerArgs),
-    mapSettlement(...handlerArgs),
-    mapCorporateActions(...handlerArgs),
-    mapProposal(...handlerArgs),
-    mapTrustedClaimIssuerTicker(...handlerArgs),
-    mapHeldTokens(eventId as EventIdEnum, moduleId as ModuleIdEnum, args),
+    mapIdentities(...handlerArgs),
+    // TODO @prashantasdeveloper remove the commented out code
+
+    // mapStakingEvent(...handlerArgs),
```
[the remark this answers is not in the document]
prashantasdeveloper: Idea is that we would migrate and uncomment each event one by one. (That is why I have also removed the tests coverage from tests/integration for now and migrated claims and identity tests to tests/entities)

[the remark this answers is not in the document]
prashantasdeveloper: I specifically asked Victor to mark sonar to won't fix

prashantasdeveloper:
@polymath-eric  This PR gets merged in `feat/wrap_up_native` which we won't merge in `develop` until 5.0.0 support PR is ready.
For updating other entities, I would again branch out of the wrap_up_native and continue the same

src/mappings/mappingHandlers.ts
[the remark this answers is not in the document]
polymath-eric: OK, that makes sense.

Can you add more detail to the comment. Something short like `These are being migrated one at a time` or something?

polymath-eric (approved):
Thanks for clarifying about why the handlers were commented.

LGTM

PR #64, opened 2022-05-31
feat: 🎸 Revamp settlements related entities

prashantasdeveloper:
### Description

- Created new entities - Portfolio, Leg, Venue and changed schema accordingly
- Added new handler to map portfolios. This handles `PortfolioCreated`, `PortfolioRenamed`,
`PortfolioCustodianChanged`, `PortfolioDeleted`, `MovedBetweenPortfolios` events
- Added event handling for `Venue` related events - `VenueCreated`, `VenueDetailsUpdated`, `VenueTypeUpdated`
- Changed the json mapping of legs to entity mapping
- Asset transfer previously handled via `InstrucionExecuted` event of `Settlement` event
is now handled by `Transfer` event of `Asset` module

### JIRA Link

DA-312

### Checklist


db/localMigration.sql
```diff
@@ -9,6 +9,10 @@
 INSERT INTO "public"."blocks" ("id", "block_id", "parent_id", "hash", "parent_hash", "state_root", "extrinsics_root", "count_extrinsics", "count_extrinsics_unsigned", "count_extrinsics_signed", "count_extrinsics_error", "count_extrinsics_success", "count_events", "datetime", "spec_version_id", "created_at", "updated_at") VALUES
 ('-1', -1, -1, '0x2', '0x6', '0xc', '0xb0', 1, 1, 0, 0, 1, 1, now(), '3000', now(), now()) ON CONFLICT(id) DO NOTHING;
 
+--Add entry for 0x0100
```
polymath-eric, line 12: I think this comment is off.

If you know them, maybe these should be "add entry for Alice", "add entry for Bob" etc. I am not sure about the 0x000 account though.

If you are not sure OK to leave as is

schema.graphql
```diff
@@ -242,18 +255,28 @@ type Settlement @entity {
   eventId: String!
   addresses: [String!]! @index(unique: false)
   result: String!
-  legs: [Leg!]!
+  legs: [Leg!]! @derivedFrom(field: "settlement")
+}
+
+type Venue @entity {
+  id: ID! # venueId
+  identity: Identity!
```
polymath-eric, line 263:
```suggestion
  owner: Identity!
```

thoughts?

src/mappings/entities/mapPortfolio.ts
```diff
@@ -0,0 +1,146 @@
...
+export const getPortfolio = async ({
+  identityId,
+  number,
+}: Pick<Portfolio, 'identityId' | 'number'>): Promise<Portfolio> => {
+  const portfolioId = `${identityId}/${number}`;
+
+  let portfolio = await Portfolio.get(portfolioId);
+
+  if (!portfolio) {
+    if (number === 0) {
```
polymath-eric, line 24: What about making a default portfolio when an Identity is created instead?

I think that would make more consistent query results since every Identity should have a portfolio even if its empty.

src/mappings/entities/mapPortfolio.ts
```diff
@@ -0,0 +1,146 @@
...
+): Promise<void> => {
+  const [, rawFromPortfolio, rawToPortfolio, rawTicker, rawAmount] = params;
+  const address = getSignerAddress(event);
+  const from = getPortfolioValue(rawFromPortfolio);
+  const to = getPortfolioValue(rawToPortfolio);
+  const ticker = serializeTicker(rawTicker);
+  const amount = getBigIntValue(rawAmount);
+
+  const settlementId = `${blockId}/${event.idx}`;
+  const settlement = Settlement.create({
```
polymath-eric, line 106: I thought a portfolio movement was different from a Settlement. Are you sure we need to make a settlement for this?

It looks like this is how it was before, but since we are changing things I am thinking we'd be able to split this up now. Seems a bit messy shoe horning it into Settlements.

I might be wrong and these actually are Settlements, but I am thinking they are different things on chain.

src/mappings/entities/mapSettlement.ts
```diff
@@ -31,175 +38,212 @@ const finalizedEvents: EventIdEnum[] = [
   EventIdEnum.InstructionFailed,
 ];
 
-// Translates events into a settlements table. This includes transfers between a users portfolio combined with completed Instructions.
-export async function mapSettlement(
+const eventResultMap = {
```
polymath-eric, line 41: Maybe

```suggestion
const instructionResultMap = {
```

tests/entities/__snapshots__/assetHolders.test.ts.snap
```diff
@@ -7,7 +7,14 @@ Object {
     "nodes": Array [
       Object {
         "__typename": "AssetHolder",
-        "amount": "101",
+        "amount": "-1000000",
+        "did": "0x0000000000000000000000000000000000000000000000000000000000000000",
```
polymath-eric, line 11: Are you sure there really is a all 0 did, and this isn't the result of serializing something that was empty?

polymath-eric (commented):
Overall looks good. I'd like to see if we can get portfolio movements separated from settlements if we are changing things.

src/mappings/entities/mapPortfolio.ts
prashantasdeveloper: I kept it as it is because in tooling while fetching settlements data `MovedBetweenPortfolios` was considered. I'm not sure we would want this, but as current use case it fetches all settlements where a DID or a specific portfolio is a part of and then also considers movement between portfolios. Ref -> https://github.com/PolymathNetwork/tooling-gql/blob/develop/src/controllers/settlement.ts

prashantasdeveloper: @VictorVicente Would like to know your thoughts here

src/mappings/entities/mapSettlement.ts
```diff
@@ -31,175 +38,212 @@ const finalizedEvents: EventIdEnum[] = [
...
+  legIndex: number,
+  ticker: string,
+  amount: bigint,
+  from: Pick<Portfolio, 'identityId' | 'number'>,
+  to: Pick<Portfolio, 'identityId' | 'number'>
+): Promise<void> => {
+  await Promise.all([getPortfolio(from), getPortfolio(to)]);
+
+  return Leg.create({
+    id: `${blockId}/${instructionId ?? 'movePortfolio'}/${legIndex}`,
```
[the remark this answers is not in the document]
prashantasdeveloper: I too don't like this but was not sure how to make it unique? Other option was to create an instruction in case of event - `MovedBetweenPortfolios`. 

What about directly - `${blockId}/${legIndex}` in this case?

tests/entities/__snapshots__/assetHolders.test.ts.snap
prashantasdeveloper: I'll have to check once again.

src/mappings/entities/mapPortfolio.ts
prashantasdeveloper: I like the idea. I'll add a default portfolio whenever a new Identity is created

schema.graphql
prashantasdeveloper: I'm OK with both. I'll change to owner

polymath-eric: I think it reads a bit easier for the person making a query.

src/mappings/entities/mapSettlement.ts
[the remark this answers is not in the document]
prashantasdeveloper: In that case i think `blockId/event.idx/legIndex` should work for sure

tests/entities/__snapshots__/assetHolders.test.ts.snap
prashantasdeveloper: Block 132 had this
Screenshot 2022-06-01 at 1 19 42 AM

db/localMigration.sql
prashantasdeveloper: I'm not sure which belongs to whom so I left 0x00, 0x01 ...

tests/entities/__snapshots__/assetHolders.test.ts.snap
polymath-eric: Thanks for checking.

I think people can put anything that looks like a DID as the target and the chain doesn't check if it exists. So we might need to think about this a bit more.

If thats the case, we might need to create a place holder identity.

Slack thread when this caused me an issue: https://polymathnetwork.slack.com/archives/C01RLA867C5/p1649454506627789?thread_ts=1649347860.107079&cid=C01RLA867C5

src/mappings/entities/mapPortfolio.ts
VictorVicente: I agree we should avoid to mix them here. We are forcing portfolio movements to fit into settlement model when they are not the same. 
I think we should track these movements in a different entity.
And the SDK should mix them when getting `portfolioTransactionHistory`. SDK could fetch them with two different calls and merge them into the same DTO by ordering them by block id to keep the same results we have right now.

WDYT @polymath-eric @prashantasdeveloper ?

polymath-eric: +1 to mixing them in the SDK. The reason they are like that is because that was what the dashboard wanted, so the SDK seems like a good place to handle that concern.

VictorVicente: Dashboard requirement makes sense. They want to show the complete history for a portfolio. That includes portfolio movements, without them balances could not be ok in that history. 
But it makes sense to track them independently in SubQuery, and return the whole history in the SDK IMO.

src/mappings/util.ts
```diff
@@ -247,3 +249,50 @@ export function addIfNotIncludes<T>(arr: T[], item: T): void {
     arr.push(item);
   }
 }
+
+type MeshPortfolio = {
```
polymath-eric, line 253: (nit) I think `interface` is preferred when possible. (LegDetails too)

If `Pick` doesn't work with interface then OK to leave, but I think it should.

polymath-eric (commented):
LGTM, one small nit about interface over type if possible.

polymath-eric (approved):

PR #65, opened 2022-06-08
feat: 🎸 Revamp all entities to support block details

prashantasdeveloper:
### Description

* Revamp all entities to support block details
* Add new `Distribution` entity
* Rename `HistoryOfPaymentEventsForCa` to `DistributionPayment`
* Drop `WithholdingTaxesOfCa` and migrate the tax calculation to `Distribution` entity
* Add mapping for `Identity` and `Asset` in all existing entities

### JIRA Link

DA-314

### Checklist


polymath-eric (commented):
LGTM, let me know when its out of draft and I'll have another look.

src/mappings/entities/mapCorporateActions.ts
```diff
@@ -1,124 +1,105 @@
...
+): Promise<void> => {
+  const [rawEventDid, , rawCaId, , rawAmount, rawTax] = params;
+
+  const targetId = getTextValue(rawEventDid);
+  const { localId, assetId } = getCaIdValue(rawCaId);
+  const amount = getBigIntValue(rawAmount);
+  const tax = getBigIntValue(rawTax);
+
+  const distribution = await Distribution.get(`${assetId}/${localId}`);
+  distribution.taxes += (amount * tax) / BigInt(1000000);
```
polymath-eric, line 62: Dividing like this can loose accuracy in the number, especially if there are a lot of small transactions. I would think its better to leave it in the smallest unit.

It might be the case that they do get rounded and this is correct though. If thats the case then I think `BigInt(1000000)` be a constant instead of a magic number.

polymath-eric (commented):
Not sure about rounding the taxes, but otherwise LGTM

src/mappings/entities/mapCorporateActions.ts
prashantasdeveloper: I wasn't sure either, then I referred back the tooling-gql and apparently it was done there. I thought it's better to leave that way. I don't have problem removing this as it could easily be handled on the SDK side

tests/entities/__snapshots__/assetHolders.test.ts.snap
```diff
@@ -7,42 +7,42 @@ Object {
     "nodes": Array [
       Object {
         "__typename": "AssetHolder",
-        "amount": "-1",
+        "amount": "-1000000",
```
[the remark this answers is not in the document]
prashantasdeveloper: Yes. Whenever a new asset is issued apparently a transfer happens from 0x00 to that DID

src/mappings/entities/mapCorporateActions.ts
polymath-eric: I think it would be better to not lose accuracy in the DB, and convert it in the SDK.

I think tooling is different since it would be storing the raw events and wouldn't compound the error over many events.

polymath-eric (approved):

PR #67, opened 2022-06-13
feat: 🎸 Map new statistics events added in 5.0.0

prashantasdeveloper:
### Description

Added a new mapper to handle statistics events - 
* StatTypesAdded
* StatTypesRemoved
* SetAssetTransferCompliance
* TransferConditionExemptionsAdded
* TransferConditionExemptionsRemoved

Following new entities have been created for the same - `StatType`, `TransferCompliance, `TransferComplianceExemption`

### JIRA Link

DA-344

### Checklist


schema.graphql
```diff
@@ -419,6 +419,43 @@ type TransferManager @entity {
...
+  createdBlock: Block!
+  updatedBlock: Block!
+}
+
+type TransferComplianceExemption @entity {
+  id: ID! #asset/opType/claimType
+  asset: Asset! @index(unique: false)
+  opType: String!
+  claimType: String
+  exemptedEntities: [String]!
```
polymath-eric, line 454: I think these are usually DIDs, maybe we could make a many too many to Identities. We'd probably have to handle the case where an exemption is set for non existing Identity with an upsert like function though.  

My concern is if this gets large we'll have no pagination. To me it doesn't seem like a field this should get embedded, even though it does make the query flatter.

Might be OK to leave as it is, but its a breaking change to address later.

polymath-eric (commented):
LGTM for the most part, left a comment around making exemptedEntities their own thing.

polymath-eric:
Seems like github is having an issue for yarn cache, which is what made your CI red I think: https://github.com/actions/cache/issues/820

prashantasdeveloper:
Yarn issue got automatically resolved by re running the jobs. Now it seems subquery isn't able to catch up in tests. I think it is the asset name change that I did (using hexToString), but i want to test it out first and polymesh-local isn't bringing up the 4.0.0 chain 
```
➜  polymesh-local git:(add-5.0.0) ./bin/run start -s 4.0.0 -v 4.0.0 --clean -o chain
Removing old state... done
No volumes detected. Initializing volumes... done
Error response from daemon: Head "https://registry-1.docker.io/v2/library/ubuntu/manifests/latest": received unexpected HTTP status: 500 Internal Server Error
Loading chain snapshot... !
    Error: Command failed: docker pull ubuntu
    Error response from daemon: Head "https://registry-1.docker.io/v2/library/ubuntu/manifests/latest": received unexpected HTTP status: 500 Internal Server Error
```
Can you check this? FYR I used your latest 5.0.0 branch

schema.graphql
prashantasdeveloper: Since we could add a non-existing entity, I have stored the `exemptedEntity` as string. I have removed the array part so that we could avoid the large values in this column and take advantage of pagination as we did for claims.

src/mappings/entities/mapAsset.ts
```diff
@@ -56,7 +58,13 @@ const handleAssetCreated = async (blockId: string, params: Codec[]): Promise<voi
   // Name isn't present on the event so we need to query storage
   // See MESH-1808 on Jira for the status on including name in the event
   const rawName = await api.query.asset.assetNames(rawTicker);
-  const name = hexToString(getTextValue(rawName));
+  const nameString = getTextValue(rawName);
+  let name;
+  if (isHex(nameString)) {
```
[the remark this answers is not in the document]
polymath-eric: thanks, that clears things up and makes it easier to remove once all the 4.0 chains are upgraded.

schema.graphql
```diff
@@ -419,6 +419,43 @@ type TransferManager @entity {
...
+  claimValue: String
+  claimIssuer: Identity
+  min: BigInt
+  max: BigInt
+  createdBlock: Block!
+  updatedBlock: Block!
+}
+
+type TransferComplianceExemption @entity {
+  id: ID! #asset/opType/claimType/exemptedEntityId
```
polymath-eric, line 450: Could this have a reference to the TransferCompliance entity?

To me it seems like it would be hard to figure out who is exempted from a condition now.

prashantasdeveloper: I don't think so we can have any relationship there. The exemption are tagged with an ExemptKey which is combination of `assetId + opType + claimType` and transfer compliance are tagged with `StatType` which may or may not depend on issuerId, which we wouldn't have in case of exempted entities. Do you have a way around this?

polymath-eric: Hmm, I think you are right. In the SDK we make it seem like the exemptions are part of the restriction itself, but on chain there isn't a relationship like that.

I guess this will have to do.

prashantasdeveloper:
It seems tests keep on running on their own terms. Previously one of my commits [8f039e6](https://github.com/PolymathNetwork/polymesh-subquery/pull/65/commits/8f039e6eb1912e4574c74b118835b48611a95c2f) failed due to staking snapshot errors and to fix it I added the exact same [snippet](https://github.com/PolymathNetwork/polymesh-subquery/pull/65/commits/87fa8ff66f9593b3f279696c4e9d1f561d6d6edb) which caused the current build to fail. Lets see removing this fixes the build or not

polymath-eric:
I see, its staking events, which are dependent on how long the chain has been running.

I think with jest we might be able to make the staking snapshot test to be a "includes these results" instead of the exact match its trying for.

polymath-eric:
Seems like there are matchers, but they don't seem to cover the what we need exactly: https://jestjs.io/docs/snapshot-testing#property-matchers

Length of the array we might be able to say "any number", but for events I am not sure how to filter them without hard coding them.


An alternative might be to make sure the polymesh-local chain doesn't publish blocks in this environment. I think thats why its inconsistent here.

schema.graphql
```diff
@@ -419,6 +420,43 @@ type TransferManager @entity {
...
+
+type TransferCompliance @entity {
+  id: ID! # ticker/op/type/claimIssuer
+  asset: Asset! @index(unique: false)
+  type: String!
+  statType: StatType!
+  # value is present only in case of MaxInvestorCount and MaxInvestorOwnership
+  value: BigInt
+  # following values are present in case of ClaimCount and ClaimOwnership
+  claimType: String
```
polymath-eric, line 441: Aren't these repeating whats in `StatType`?

schema.graphql
```diff
@@ -419,6 +420,43 @@ type TransferManager @entity {
...
+  updatedBlock: Block!
+}
+
+type TransferCompliance @entity {
+  id: ID! # ticker/op/type/claimIssuer
+  asset: Asset! @index(unique: false)
+  type: String!
+  statType: StatType!
+  # value is present only in case of MaxInvestorCount and MaxInvestorOwnership
+  value: BigInt
```
polymath-eric, line 439: I think it would be nicer to have `value` as something like `BigInt | StatType`, vs having optional fields.

Basically when its a "scoped claim", the "value" is a tuple (StatClaim, issuer, min, max?), I think we are capturing this in StatType, so do we need the `claim...` fields in here as well?

Either value is present, or if claim stuff is there, then 3/4 fields are not optional

polymath-eric (commented):
Left a comment around `TransferCompliance` model, I think we can improve it a bit. 

I am looking at it in the SDK, so let me know if you have questions.

schema.graphql
polymath-eric: I think the flat values will work, but add some extra confusion for the consumer

prashantasdeveloper: I don't think we can provide optional types like `BigInt | StatType`. I'll keep the flat structure for now. 

To avoid confusion what we can do is rename `value` to `countValue` and `min, max` as `minClaimPercentage, maxClaimPercentage` or `minClaimBalance, maxClaimBalance` . Maybe that would help?

schema.graphql
prashantasdeveloper: Originally I didn't have the statType here but when I tried out some transactions apparently we can't delete a statType if transfer compliance is mapped to it, so I added it. I'm OK with removing it, but now I think it would be confusing to consumers to interpret if some values are present in stat type and some in transfer compliance. It already has a confusing flat structure and on top of that having some values come in nested format, I'm not a big fan of it. But said that I can move it if you say so.

schema.graphql
polymath-eric: I think having the names spelled out would make it clearer.

`maxClaimCount` is optional, but `maxClaimOwnership` isn't btw.

schema.graphql
```diff
@@ -419,6 +420,43 @@ type TransferManager @entity {
   updatedBlock: Block!
 }
 
+type StatType @entity {
+  id: ID! # ticker/op/type/claimIssuer
+  asset: Asset!
+  opType: String!
+  claimType: String
```
polymath-eric, line 427: Does SubQuery support enums with the new version? I remember Diego submitted a PR upstream for it but not sure if it got merged in.

If it does, then I think it would be nice to have this as an enum type.

If not OK to leave as string though.

schema.graphql
polymath-eric: Makes sense, generally I don't like duplicated data, but probably OK to leave in. Maybe the claim values get grouped in a JSON field? `claim: {value: String!, type: String!, issuer: Identity}`?

schema.graphql
prashantasdeveloper: Introducing all named variables means a lot of variables, hence i stuck with min and max here, so that consumer can deduce what it means based on other values (type, statType). I think for now it should work.

schema.graphql
prashantasdeveloper: I checked and Yes enums are working. I'll create a separate task to add enum support wherever possible in subquery

schema.graphql
prashantasdeveloper: One problem with the JSON you specified is that we can't use an entity within a JSON. So the issuer field then has to be converted to string

schema.graphql
polymath-eric: Makes sense

src/mappings/entities/mapSettlement.ts
```diff
@@ -48,34 +48,40 @@ const instructionResultMap = {
...
   instructionId: string,
-  settlementId: string
-): Promise<void> => {
+  settlementId?: string
+): Promise<Promise<void>[]> => {
   const legs = await Leg.getByInstructionId(instructionId);
 
-  await Promise.all(
-    legs.map(leg => {
+  return legs.map(leg => {
```
polymath-eric, line 75: If this did  `return Promise.all(legs.map(leg => ....)` I think the function return type would be `Promise<void[]>` which looks neater than having a nested promise

PR #68, opened 2022-07-06
fix: 🐛 serialize ticker for type `PolymeshPrimitivesTicker`

prashantasdeveloper:
### Description

Type value for `Ticker` was changed to `PolymeshPrimitivesTicker` with 5.0.0 upgrade and subquery missed serialising objects for this type. Added the check for the new type as well along with the old type.

### JIRA Link

DA-371

### Checklist


polymath-eric (approved):

polymath-eric:
Do lines like
```
else if (rawType === 'AccountId')`
```

also need to get updated?

prashantasdeveloper:
Other check in that file are working fine. I cross checked the availability of those in 5.0.0 specBundles and those are present.

PR #73, opened 2022-08-09
feat: 🎸 Add inserts for predefined DIDs

prashantasdeveloper:
### Description

Manages migration queries to be run for SubQuery. There a number of SystematicIssuers (fixed identities created in genesis block) in all the environments that should be added in subquery initial migration. These identities are defined here: https://github.com/PolymeshAssociation/Polymesh/blob/d45fd1a161990310242f21230ac8a1a5d15498eb/pallets/common/src/constants.rs#L23

Adds a method to generate inserts for genesisBlock, all predefined DIDs and their respective default portfolios.

Also, removes handling of Asset transfers for null DID `0x000...00000`

### JIRA Link

[DA-394](https://polymath.atlassian.net/browse/DA-394)

### Checklist


db/migration.ts
```diff
@@ -0,0 +1,48 @@
...
+    '0x73797374656d3a706f6c796d6174685f636c61737369635f6d69670000000000', // Classic Migration
+    '0x73797374656d3a666961745f7469636b6572735f7265736572766174696f6e00', // FIAT Tickers Reservation
+    '0x73797374656d3a726577617264735f6d6f64756c655f64696400000000000000', // Rewards
+  ];
+
+  // We also define identities for first 16 GC identities
+  const gcIdentities = Array(16)
+    .fill('')
+    .map((v, index) => {
+      const twoDigitNumber = `0${index + 1}`.slice(-2);
```
polymath-eric, line 28: There is `padStart` method as well

`${index+1}.padStart(2, '0')`

OK to leave if you prefer though

polymath-eric (approved):

db/migration.ts
prashantasdeveloper: I was OK with both. Made the change

polymath-eric (approved):

PR #74, opened 2022-08-11
fix: 🐛 Handle camelCase for structure keys and add deprecated enum values

prashantasdeveloper:
### Description

The claim events on the chain were having varied casing for attribute names. 

For example, wrt staging environment, block [56926](https://polkadot.js.org/apps/#/explorer/query/56296) was having snake case for attributes as below - 

Screenshot 2022-08-11 at 7 05 24 PM

But for the block [923906](https://polkadot.js.org/apps/#/explorer/query/923906) (, attribute names were camelCase as shown below - 

Screenshot 2022-08-11 at 7 06 44 PM

It seems this is after the v14 metadata. Polkadot says - 

> Be aware that in the JS version naming defaults to camelCase where names of fields in Substrate defaults to snake_case. (Each version aligning with conventions in the respective languages)

**Changelog** 
* This change makes sure that while fetching any attribute values from event both cases are checked for to support backward compatibility with old blocks from 4.0.0 node as well
* Also adds deprecated eventId , moduleId and callId enums

### Checklist

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -49,6 +49,13 @@ enum ModuleIdEnum {
   polymeshcontracts
   preimage
   testutils
+
+  ## deprecated ##
+  confidential @deprecated
+  dividend @deprecated
+  exemption @deprecated
+  finalitytracker @deprecated
+  stocapped @deprecated
 }
 
 enum EventIdEnum {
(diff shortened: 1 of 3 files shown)


polymath-eric (approved):
lgmt

VictorVicente (approved):

PR #77, opened 2022-08-15
fix: 🐛 Fix `extractValue` method to handle `0` value

prashantasdeveloper:
### Description

`extractValue` method was returning undefined for key value `0`. (Found out via block [30594](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstaging-rpc.polymesh.live#/explorer/query/30594) in staging). 

Fixed the logic to return 0

### Checklist

### DIFF
--- a/src/mappings/generatedColumns.ts
+++ b/src/mappings/generatedColumns.ts
@@ -41,7 +41,10 @@ export const extractEventArgs = (args: any[]) => {
  */
 export const extractValue = (obj: unknown, key: string): string => {
   if (obj) {
-    return obj[key] || obj[snakeToCamelCase(key)];
+    if (Object.keys(obj).includes(key)) {
+      return obj[key];
+    }
+    return obj[snakeToCamelCase(key)];
   }
   return undefined;
 };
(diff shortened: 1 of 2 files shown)


src/mappings/generatedColumns.ts
```diff
@@ -41,7 +41,10 @@ export const extractEventArgs = (args: any[]) => {
  */
 export const extractValue = (obj: unknown, key: string): string => {
   if (obj) {
-    return obj[key] || obj[snakeToCamelCase(key)];
+    if (Object.keys(obj).includes(key)) {
```
polymath-eric, line 44:
```suggestion
    if (key in obj) {
```
JS has the `in` operator. Would that work?

polymath-eric (commented):
one comment to potentially clean up the syntax

src/mappings/generatedColumns.ts
prashantasdeveloper: I tried but it wasn't working so I switched to using keys

polymath-eric (approved):

PR #78, opened 2022-08-15
Catch and log errors

polymath-eric:
### Description

Adds some error logging based on this slack message: https://polymathnetwork.slack.com/archives/C01CBL975KN/p1660589054279759?thread_ts=1660330334.658039&cid=C01CBL975KN

I didn't add in an env flag since its already with the logger which has its own levels to set.

### Checklist


docker-compose.yml
```diff
@@ -44,8 +44,6 @@ services:
       # Mainnet
       # NETWORK_ENDPOINT: wss://mainnet-rpc.polymesh.network
       # NETWORK_CHAIN_ID: '0x6fbd74e5e1d0a61d52ccfe9d4adaed16dd3a7caa37c6bc4d0c2fa12e8b2f4063'
-    volumes:
```
polymath-eric, line 47: This was causing issues for some people with file permission issues. Its not needed since the image copies over what it needs when we build it.

src/mappings/util.ts
```diff
@@ -258,7 +258,7 @@ export const getExemptKeyValue = (
 };
 
 export const getExemptionsValue = (exemptions: Codec): string[] => {
-  return JSON.parse(exemptions.toString()) || [];
+  return exemptions.toJSON() as string[];
```
prashantasdeveloper, line 261: This was the reason for error in block `384889` 
```
Received an error in handleEvent function: SyntaxError: Unexpected token x in JSON at position 2 
```

VictorVicente (approved):

docker-entrypoint.sh
```diff
@@ -27,7 +27,10 @@ envsubst <project.template.yaml> project.yaml
 
 (npm run sql || (sleep 3 && kill "$$")) &
 
-node --max-old-space-size=1536 \
+# Allow configuring node memory. Default to 1.5MB, should be ~75% of available RAM
+NODE_SPACE=${MAX_OLD_SPACE_SIZE:-1536}
```
prashantasdeveloper, line 31: Shouldn't we specify this var in docker-compose so that user can understand this can be configured? Also, a line in README would be beneficial. Can you please add that?

prashantasdeveloper: Maybe its better to mention in comment in the README

prashantasdeveloper (approved):
Left a nit comment, other than that LGTM

docker-entrypoint.sh
polymath-eric: I was thinking of keeping this secret since most users shouldn't need to worry about it. the default of 1536 should be good enough.

Might be good to have a section with "advanced" and put it in there

PR #80, opened 2022-08-17
fix: 🐛 Correct parsing of address for secondary accounts and update Authorization entity

prashantasdeveloper:
### Description

* Schema for secondary keys was changed for 5.x.x chain due to which block #925160 was failing in staging. 
`signer: Signatory` is replaced by `key: AccountId`

* Fixed the `diff_versions` script and updated them

* Replaced `to: Identity` to `toId: String` in `Authorization` schema

### Checklist

### DIFF
--- a/src/mappings/entities/mapIdentities.ts
+++ b/src/mappings/entities/mapIdentities.ts
@@ -185,9 +185,16 @@ const handleSecondaryKeysPermissionsUpdated = async (
 ): Promise<void> => {
   const [, rawSignerDetails, , rawUpdatedPermissions] = params;
 
-  const {
-    signer: { account: address },
-  } = JSON.parse(rawSignerDetails.toString());
+  let address;
+  try {
+    // for chain version < 5.0.0
+    ({
+      signer: { account: address },
+    } = JSON.parse(rawSignerDetails.toString()));
+  } catch (_) {
+    // for chain version >= 5.0.0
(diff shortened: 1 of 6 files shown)


src/mappings/entities/mapIdentities.ts
```diff
@@ -185,9 +185,16 @@ const handleSecondaryKeysPermissionsUpdated = async (
 ): Promise<void> => {
   const [, rawSignerDetails, , rawUpdatedPermissions] = params;
 
-  const {
-    signer: { account: address },
-  } = JSON.parse(rawSignerDetails.toString());
+  let address;
+  try {
```
polymath-eric, line 189: could this be some kind of check instead of relying on an exception to be thrown? I'd prefer `if` over exception for control flow.

If there is a reason we have to use try / catch its OK, but I think its better to avoid if we can help it

polymath-eric (commented):
LGTM, left a nitty comment around try / catch block

src/mappings/entities/mapIdentities.ts
prashantasdeveloper: I'll look for a work around

prashantasdeveloper: Updated the logic using instanceof

polymath-eric (approved):

PR #82, opened 2022-08-19
chore: 🤖 Support unvalidated DID while creating legs

prashantasdeveloper:
### Description

* While creating instructions, receiver DID/Portfolio isn't validated. To support this scenario, SQ inserts apt portfolio which isn't present (without any name). For this if the underlying DID is also not present, it creates one without any primaryAccount(sets the value to `''`). 
This was found in mainnet block - [1375280](https://polymesh.subscan.io/block/1375280?tab=event)

* Fix parsing of Asset compliances for `AssetComplianceReplaced` event.

### Breaking Changes

N/A

### JIRA Link

[DA-253](https://polymath.atlassian.net/browse/DA-253)

### Checklist

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1160,14 +1160,16 @@ type Account @entity {
   updatedBlock: Block!
 }
 
+# Identity will have null values for `primaryAccount`, `eventId`, `datetime` in cases where Identity is referenced without actually being created as in Authorization/Instruction
+# Once Identity is created, values will be updated for these attributes
 type Identity @entity {
   id: ID!
   did: String! @index(unique: true)
-  primaryAccount: String! @index(unique: false)
+  primaryAccount: String @index(unique: false)
   secondaryAccounts: [Account!]! @derivedFrom(field: "identity")
   secondaryKeysFrozen: Boolean!
-  eventId: EventIdEnum!
-  datetime: Date!
+  eventId: EventIdEnum
+  datetime: Date
   createdBlock: Block!
   updatedBlock: Block!
   heldAssets: [AssetHolder]! @derivedFrom(field: "identity")
@@ -1313,6 +1315,7 @@ type PortfolioJson @jsonField {
   number: Int!
 }
 
+# For each entry in DID, a default Portfolio (which contains number as 0) entry is created.
 type Portfolio @entity {
   id: ID! # did/number
   identity: Identity! @index(unique: false)
--- a/schema.graphql
+++ b/schema.graphql
@@ -1160,8 +1160,6 @@ type Account @entity {
   updatedBlock: Block!
 }
 
-# Identity will have null values for `primaryAccount`, `eventId`, `datetime` in cases where Identity is referenced without actually being created as in Authorization/Instruction
-# Once Identity is created, values will be updated for these attributes
 type Identity @entity {
   id: ID!
   did: String! @index(unique: true)
@@ -1315,7 +1313,7 @@ type PortfolioJson @jsonField {
   number: Int!
 }
 
-# For each entry in DID, a default Portfolio (which contains number as 0) entry is created.
+# For each entry in `Identity`, a default Portfolio (which contains number as 0) entry is created.
 type Portfolio @entity {
   id: ID! # did/number
   identity: Identity! @index(unique: false)
--- a/schema.graphql
+++ b/schema.graphql
@@ -1165,11 +1165,11 @@ type Account @entity {
 type Identity @entity {
   id: ID!
   did: String! @index(unique: true)
-  primaryAccount: String @index(unique: false)
+  primaryAccount: String! @index(unique: false)
   secondaryAccounts: [Account!]! @derivedFrom(field: "identity")
   secondaryKeysFrozen: Boolean!
-  eventId: EventIdEnum
-  datetime: Date
+  eventId: EventIdEnum!
+  datetime: Date!
   createdBlock: Block!
   updatedBlock: Block!
   heldAssets: [AssetHolder]! @derivedFrom(field: "identity")
--- a/src/mappings/entities/mapIdentities.ts
+++ b/src/mappings/entities/mapIdentities.ts
@@ -11,7 +11,7 @@ import {
 } from '../../types';
 import { getTextValue } from '../util';
 import { HandlerArgs } from './common';
-import { createPortfolio } from './mapPortfolio';
+import { createPortfolio, getPortfolio } from './mapPortfolio';
 
 /**
  * Subscribes to the Identities related events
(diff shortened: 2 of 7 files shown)


src/mappings/entities/mapSettlement.ts
```diff
@@ -53,8 +54,14 @@ export const createLeg = async (
   address: string,
   legIndex: number,
   { ticker, amount, from, to }: LegDetails
-): Promise<void> =>
-  Leg.create({
+): Promise<void> => {
+  // since an instruction leg can be created without a valid DID, we make sure DB has an entry for Identity to avoid foreign key constraint
```
polymath-eric, line 58: I think we should make the reference nullable instead of making an Identity that doesn't really exist.

DID's are quasi-random, so its very unlikely someone predicted their counter party. If this happens its likely to be bad user input.

I think we should keep the shortcoming of the Leg to the leg, and not let it weaken the Identity model. I think having two fields one a nullable foreign reference is better than inserting Identities that don't really exist.

polymath-eric (changes requested):

schema.graphql
```diff
@@ -1160,14 +1160,16 @@ type Account @entity {
   updatedBlock: Block!
 }
 
+# Identity will have null values for `primaryAccount`, `eventId`, `datetime` in cases where Identity is referenced without actually being created as in Authorization/Instruction
+# Once Identity is created, values will be updated for these attributes
 type Identity @entity {
   id: ID!
   did: String! @index(unique: true)
-  primaryAccount: String! @index(unique: false)
+  primaryAccount: String @index(unique: false)
```
polymath-eric, line 1168: This will be really annoying for every user of Identity (`id.?primaryAccount`). I think this should remain `String!` since this is something the chain ensures.

polymath-eric (approved):

schema.graphql
```diff
@@ -1160,14 +1160,16 @@ type Account @entity {
...
   secondaryAccounts: [Account!]! @derivedFrom(field: "identity")
   secondaryKeysFrozen: Boolean!
-  eventId: EventIdEnum!
-  datetime: Date!
```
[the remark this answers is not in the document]
prashantasdeveloper: We can pass that along as we are passing the blockId

polymath-eric:
We talked on Slack. The idea is to set primaryAccount to be '' and enforce not null.

Although this might lead to some weird behavior, it will only be a result of incorrectly formed instructions, and will be limited in scope. This lets us keep the schema mostly as is and avoid imposing`.?` checks on queriers.

src/mappings/entities/mapIdentities.ts
```diff
@@ -63,6 +66,35 @@ const getIdentity = async (param: Codec): Promise<Identity> => {
   return identity;
 };
 
+/**
+ * Returns Identity for a given DID. If no Identity exists, it creates one along with its default Portfolio before returning the created Identity
+ */
+export const getOrCreateIdentity = async (did: string, blockId: string): Promise<Identity> => {
```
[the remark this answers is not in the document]
prashantasdeveloper: My idea is not to use this method at all places, but very selectively. I would keep the migration file for now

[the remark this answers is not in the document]
polymath-eric: Can we put a warning in the JS docs. Is `@warning` or a thing? Even a `@note`.

prashantasdeveloper:
Along with unvalidated Identity, Portfolio isn't also validated when created via webapp. So I added the logic to handle this case well. And made the discussed changes along with it

schema.graphql
```diff
@@ -1160,6 +1160,8 @@ type Account @entity {
   updatedBlock: Block!
 }
 
+# Identity will have null values for `primaryAccount`, `eventId`, `datetime` in cases where Identity is referenced without actually being created as in Authorization/Instruction
+# Once Identity is created, values will be updated for these attributes
```
polymath-eric, line 1164:
```suggestion
```

schema.graphql
```diff
@@ -1313,6 +1315,7 @@ type PortfolioJson @jsonField {
   number: Int!
 }
 
+# For each entry in DID, a default Portfolio (which contains number as 0) entry is created.
```
polymath-eric, line 1318: (nit)
```suggestion
# For each entry in `Identity`, a default Portfolio (which contains number as 0) entry is created.
```

Might be OK to leave out this comment as well.

src/mappings/entities/mapPortfolio.ts
```diff
@@ -39,6 +40,30 @@ export const createPortfolio = (
   }).save();
 };
 
+/**
+ * Creates a Portfolio if not present.
+ * This can be need in case where someone creates an Instruction before the recipient Portfolio is created
```
polymath-eric, line 45:
```suggestion
 * @note - WARNING: This is needed when an Instruction is created with a target Portfolio that doesn't exist. It should not be used unless necessary.`

File: src/mappings/entities/mapPortfolio.ts [hunk 8]
```diff
@@ -39,6 +40,30 @@ export const createPortfolio = (
...
+ */
+export const createPortfolioIfNotExists = async (
+  { identityId, number }: Pick<Portfolio, 'identityId' | 'number'>,
+  blockId: string,
+  event: SubstrateEvent
+): Promise<void> => {
+  await createIdentityIfNotExists(identityId, blockId, event);
```
polymath-eric, line 52: What about inlining this function so people don't use it by mistake? We'd be able to batch some requests as well.

src/mappings/entities/mapPortfolio.ts
```diff
@@ -50,15 +75,26 @@ const handlePortfolioCreated = async (
   const number = getNumberValue(rawPortfolioNumber);
   const name = getTextValue(rawName);
 
-  await createPortfolio(
-    {
-      identityId: ownerId,
-      number,
+  const portfolio = await Portfolio.get(`${ownerId}/${number}`);
```
polymath-eric, line 78: Shoudln't this check go into the `createPortfolioIfNotExists` function, and then we wouldn't need this check here?

If the event is `PortfolioCreated` we shouldn't have it in the DB.

src/mappings/entities/mapPortfolio.ts
```diff
@@ -39,6 +40,30 @@ export const createPortfolio = (
...
+ * This can be need in case where someone creates an Instruction before the recipient Portfolio is created
+ */
+export const createPortfolioIfNotExists = async (
+  { identityId, number }: Pick<Portfolio, 'identityId' | 'number'>,
+  blockId: string,
+  event: SubstrateEvent
+): Promise<void> => {
+  await createIdentityIfNotExists(identityId, blockId, event);
+
+  if (number) {
```
polymath-eric, line 54: Don't we need to check if it exists first?

src/mappings/entities/mapIdentities.ts
```diff
@@ -63,6 +67,40 @@ const getIdentity = async (param: Codec): Promise<Identity> => {
   return identity;
 };
 
+/**
+ * Creates an Identity if already not present. It also creates default Portfolio for that Identity
```
polymath-eric, line 71: Can we add a note that tries to dissuade people from using this function?

src/mappings/entities/mapPortfolio.ts
polymath-eric: People can send to Portofolios that don't exist? Maybe we do need this check after all.

src/mappings/entities/mapPortfolio.ts
polymath-eric: (might be better to leave as is though)

prashantasdeveloper: Inlining would mean to handle `Identity` entity changes here which I have tried to avoid. My idea was to keep the respective entity operations in its respective mapping handler (like all `Portfolio.get`, `Portfolio.create`, `Portfolio.remove` operations would be in `mapPortfolios` and we would have exported members if something is needed elsewhere)

src/mappings/entities/mapPortfolio.ts
prashantasdeveloper: > Shoudln't this check go into the createPortfolioIfNotExists function

Yes you are right. That would also prevent the method being misused for cases where Portfolio actually existed

> If the event is PortfolioCreated we shouldn't have it in the DB.

For that I added the check that in case Portfolio got inserted because of Instruction created event before being created, we can then just update the Portfolio name and updatedBlockId

> People can send to Portofolios that don't exist? Maybe we do need this check after all.

This is even more likely to happen instead of DID not being present. Like lets say if you and I had a discussion where I was sending you some ACME tokens and you said that send it to my Portfolio 3 (but that didn't existed). So I created the instruction and then you realised I need to create it to receive this. This would be handled now as well.

src/mappings/entities/mapPortfolio.ts
prashantasdeveloper: Added the check

polymath-eric (approved):

PR #83, opened 2022-08-23
fix: 🐛 Handle ticker serialization with metadata v14

prashantasdeveloper:
#### Problem 

With latest SQ sync, ticker values are not properly serialized and are getting stored in hex format.

#### Changelog

- The `serializeLikeHarvester` function wasn't able to return the serialized values for tickers since the meta-type was being received as `[u8;12]` instead of `Ticker`. To get the exact type name, `meta.fields` is now being used.
- Fixed the parsing issue of `offering_asset` for `sto.FundraiserCreated` event. [Staging block - [1765423](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstaging-rpc.polymesh.live#/explorer/query/1765423)]
- Handles missing `fromId` for `Authorization` when `identity.cdd_register_did` is invoked. This particular case occurs
 when secondary keys are also passed as params along with target account. [Mainnet block - [2277507](https://polymesh.subscan.io/block/2277507)]

### Breaking Changes

NA

### JIRA Link

[DA-253](https://polymath.atlassian.net/browse/DA-253)

### Checklist

### DIFF
--- a/src/mappings/mappingHandlers.ts
+++ b/src/mappings/mappingHandlers.ts
@@ -105,9 +105,21 @@ export async function handleToolingEvent(event: SubstrateEvent): Promise<void> {
   const eventId = event.event.method;
   const args = event.event.data.toArray();
 
-  const harvesterLikeArgs = args.map((arg, i) => ({
-    value: serializeLikeHarvester(arg, event.event.meta.args[i].toString(), logFoundType),
-  }));
+  const harvesterLikeArgs = args.map((arg, i) => {
+    let type;
+    const typeName = event.event.meta.fields[i].typeName;
+    if (typeName.isSome) {
+      // for metadata >= v14
+      type = typeName.unwrap().toString();
+    } else {
+      // for metadata < v14
+      type = event.event.meta.args[i].toString();
+    }
+    return {
+      value: serializeLikeHarvester(arg, type, logFoundType),
+    };
+  });
+
   const { eventArg_0, eventArg_1, eventArg_2, eventArg_3 } = extractEventArgs(harvesterLikeArgs);
 
   const { claimExpiry, claimIssuer, claimScope, claimType } = extractClaimInfo(harvesterLikeArgs);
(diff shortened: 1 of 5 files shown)


src/mappings/mappingHandlers.ts
```diff
@@ -105,9 +105,21 @@ export async function handleToolingEvent(event: SubstrateEvent): Promise<void> {
...
-  }));
+  const harvesterLikeArgs = args.map((arg, i) => {
+    let type;
+    const typeName = event.event.meta.fields[i].typeName;
+    if (typeName.isSome) {
+      // for metadata >= v14
+      type = typeName.unwrap().toString();
+    } else {
+      // for metadata < v14
+      type = event.event.meta.args[i].toString();
```
polymath-eric, line 116: I thought the bad values were from old events. Isn't this the same as it was before for metadata < 14?

src/mappings/mappingHandlers.ts
```diff
@@ -105,9 +105,21 @@ export async function handleToolingEvent(event: SubstrateEvent): Promise<void> {
   const eventId = event.event.method;
   const args = event.event.data.toArray();
 
-  const harvesterLikeArgs = args.map((arg, i) => ({
-    value: serializeLikeHarvester(arg, event.event.meta.args[i].toString(), logFoundType),
-  }));
+  const harvesterLikeArgs = args.map((arg, i) => {
+    let type;
+    const typeName = event.event.meta.fields[i].typeName;
```
polymath-eric, line 110: (nit) could this be in the if block?

src/mappings/mappingHandlers.ts
prashantasdeveloper: No, it isn't the same after the upgrade. I added the old vs new values here - https://polymathnetwork.slack.com/archives/C026L2FKT25/p1661246441250979?thread_ts=1661203792.561179&cid=C026L2FKT25

src/mappings/mappingHandlers.ts
prashantasdeveloper: I'm not sure what you mean here?

polymath-eric: My bad, I misread this.

polymath-eric (approved):

PR #85, opened 2022-09-09
fix: 🐛 Handle identity state when claim added

prashantasdeveloper:
### Description

With testnet block [1746733](https://polymesh-testnet.subscan.io/block/1746733?tab=event), claim added event was triggered with a non-existing target DID. Since the target claim isn't validated before depositing the event, this PR adds a check to make sure the identity exists.

### Breaking Changes

NA

### JIRA Link

DA-253

### Checklist


polymath-eric (approved):

prashantasdeveloper:
Based on the discussion [here](https://polymathnetwork.slack.com/archives/C01CBL975KN/p1662726588004639) we would need another entity/attribute to be created. But as a quick solution this should do the trick.

PR #86, opened 2022-09-11
fix: 🐛 Testnet native graphql sync issues

prashantasdeveloper:
### Description

Adds a `deletedAt` attribute to `Portfolio` entity which now gets populated when a portfolio is deleted, instead of removing the entity. This is to prevent the deletion of all other entities referencing the deleted Portfolio

### JIRA Link

DA-253

### Checklist

### DIFF
--- a/src/mappings/entities/mapIdentities.ts
+++ b/src/mappings/entities/mapIdentities.ts
@@ -267,14 +267,26 @@ const handleSecondaryKeysPermissionsUpdated = async (
   await permissions.save();
 };
 
+type MeshAccount = string | { account: string };
+
 const handleSecondaryKeysRemoved = async (params: Codec[]): Promise<void> => {
   const [, rawAccounts] = params;
 
-  const accounts = JSON.parse(rawAccounts.toString());
+  const accounts = rawAccounts.toJSON() as MeshAccount[];
 
-  const removePromises = accounts.map(({ account }) => Account.remove(account));
+  const removePromises = accounts.map(account => {
+    let address;
+    if (typeof account === 'string') {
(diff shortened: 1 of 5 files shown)


polymath-eric (approved):

src/mappings/entities/mapIdentities.ts
```diff
@@ -267,14 +267,26 @@ const handleSecondaryKeysPermissionsUpdated = async (
...
+
 const handleSecondaryKeysRemoved = async (params: Codec[]): Promise<void> => {
   const [, rawAccounts] = params;
 
-  const accounts = JSON.parse(rawAccounts.toString());
+  const accounts = rawAccounts.toJSON() as MeshAccount[];
 
-  const removePromises = accounts.map(({ account }) => Account.remove(account));
+  const removePromises = accounts.map(account => {
+    let address;
```
polymath-eric, line 278: (style/optional) What do you think about defining a util function `accountToAddress(account: Account | string): string` that contains this logic. Then call map two times (or one map and call it),

```ts
accounts
  .map(accountToAddress)
  .map(address => [Account.remove(address), Permissions.remove(address)]
 ```

The above is double iterating, but you can do it in one map by calling the function directly too.


Its OK to leave since its not duplicated, but I suspect there will be in the future.

polymath-eric (approved):
left a comment, but ok to merge with or without it

src/mappings/entities/mapIdentities.ts
prashantasdeveloper: If you see the `mapIdentities`, all have different account parsing logic. That's why I didn't extract the logic.

PR #89, opened 2022-09-29
feat: 🎸 add memo field to instruction

polymath-eric:
### Description

this is needed for chains >= 5.1.0 as they may have memo in InstructionCreatedEvent

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-424](https://polymesh.atlassian.net/browse/DA-424)

### Checklist


src/mappings/entities/mapAsset.ts
```diff
@@ -55,20 +55,20 @@ const handleAssetCreated = async (
...
-  const [rawOwnerDid, rawTicker, divisible, rawType, , disableIu] = params;
+  const [rawOwnerDid, rawTicker, divisible, rawType, , disableIu, rawAssetName] = params;
   const ownerId = getTextValue(rawOwnerDid);
   const ticker = serializeTicker(rawTicker);
   const type = getTextValue(rawType);
   /**
-   * Name isn't present on the event so we need to query storage.
-   * See MESH-1808 on Jira for the status on including name in the event
+   * Name isn't present on the old events so we need to query storage.
+   * Events from chain >= 5.1.0 has it, so its faster to sync using it
```
polymath-eric, line 64:
```suggestion
   * Events from chain >= 5.1.0 has it, and its faster to sync using it
```

VictorVicente (commented):
Is it required a reset? Or we can solve it by a migration?

src/mappings/entities/mapSettlement.ts
```diff
@@ -146,11 +146,20 @@ const handleInstructionCreated = async (
...
+    rawSettlementType,
+    rawTradeDate,
+    rawValueDate,
+    rawLegs,
+    rawOptMemo,
+  ] = params;
 
   const legs = getLegsValue(rawLegs);
   const instructionId = getTextValue(rawInstructionId);
+  const memo = getTextValue(rawOptMemo);
```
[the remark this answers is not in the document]
polymath-eric: null should be returned, but that is what is causing the test to fail. Lets see if my fix works

polymath-eric:
We should be able to add memo onto instruction table to avoid a resync, since all previous instructions won't have one

We should still resync just in case though

polymath-eric:
Actually I think we had issues with adding enums in before. We might need to resync after all

src/mappings/util.ts
```diff
@@ -121,15 +121,15 @@ export const getFirstValueFromJson = (item: Codec): string => {
...
+  return item?.toString().trim().length > 0 ? item.toString().trim() : null;
 };
 
 export const getNumberValue = (item: Codec): number => {
   return Number(getTextValue(item));
 };
 
 export const getDateValue = (item: Codec): Date => {
-  return item.toString().trim().length === 0 ? null : new Date(Number(item.toString()));
+  return item.toString().trim().length > 0 ? new Date(Number(item.toString())) : null;
```
polymath-eric, line 132: This was fine, but I changed it so it would look the same, and if it were changed it wouldn't be the same bug as getTextValue had.

polymath-eric:
The migration to avoid a resync should be:

```sql
ALTER TYPE *CallIdEnum* ADD VALUE 'redeem_from_portfolio' AFTER 'redeem';
ALTER TYPE *CallIdEnum* ADD VALUE 'add_instruction_with_memo' AFTER 'add_instruction';
ALTER TYPE *CallIdEnum* ADD VALUE 'add_and_affirm_instruction_with_memo' AFTER 'add_and_affirm_instruction';

ALTER TABLE instructions ADD memo TEXT;
```

SubQuery gives random names to enums internally though so you need to figure it out by looking at the enum:
Heres how with psql, but graphical interfaces should expose them as well:

 \dT to show enums, the description will contain what they are in GraphQL example: public_enum_0bf3c7d4ef @enumName CallIdEnum

VictorVicente (approved):

PR #93, opened 2022-10-25
chore: 🤖 Add support for schema migration

prashantasdeveloper:
### Description

This allows to specify the previous and current version of subquery node being run. If any schema migrations are needed, this runs the migration on startup.

### Breaking Changes

NA

### JIRA Link

DA-436

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,4 @@ More advanced options are:
 
 - `NO_NATIVE_GRAPHQL_DATA` — this will set the event handler to only record what is necessary for tooling gql. This allows a stable indexer to be ran as native GraphQL handlers get developed
 - `MAX_OLD_SPACE_SIZE` — this will be passed onto the node process as `--max-old-space-size` flag. The recommendation is for this to be ~75% of available RAM. Defaults to 1536, a setting appropriate for 2GB.
+- `CURRENT_SQ_VERSION` & `PREVIOUS_SQ_VERSION` - this will automatically run the required migration files to migrate your database, if any, from previously run subquery version to current one. This can be used to avoid resync and incorporate schema changes required for current version.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,9 @@ services:
       DB_HOST: postgres
       DB_PORT: 5432
       START_BLOCK: 1
+      ## Set the below two values to run all migration required from version `PREVIOUS_SQ_VERSION` to `CURRENT_SQ_VERSION`
+      PREVIOUS_SQ_VERSION: 5.3.8
+      CURRENT_SQ_VERSION: 5.4.3
       # NO_NATIVE_GRAPHQL_DATA: 'TRUE' # Setting this value only processes tooling specific handlers. Native subquery entities are ignored.
       # Polymesh-local
       NETWORK_ENDPOINT: ws://host.docker.internal:9944
--- a/scripts/run-sql.ts
+++ b/scripts/run-sql.ts
@@ -1,6 +1,6 @@
-import { createConnection } from 'typeorm';
+import { Connection, createConnection } from 'typeorm';
 import { env, chdir } from 'process';
-import { readFileSync } from 'fs';
+import { readFileSync, readdirSync } from 'fs';
 import { migrationQueries } from '../db/migration';
 chdir(__dirname);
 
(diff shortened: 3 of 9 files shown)


docker-compose.yml
```diff
@@ -30,6 +30,9 @@ services:
       DB_HOST: postgres
       DB_PORT: 5432
       START_BLOCK: 1
+      ## Set the below two values to run all migration required from version `PREVIOUS_SQ_VERSION` to `CURRENT_SQ_VERSION`
+      PREVIOUS_SQ_VERSION: 5.3.8
```
polymath-eric, line 34: I'd prefer if the user could just start the new version and have these be automatically ran with config to enable auto migration. (maybe default to auto migrate?). The user has already specified the versions so this seems like double input that will be error prone.

SQ has sequlize as a dependency, so we should be able to use that migration system: https://sequelize.org/docs/v6/other-topics/migrations/

The CLI should be scriptable and ran with the one making tooling-gql style tables.

scripts/run-sql.ts
```diff
@@ -1,6 +1,6 @@
-import { createConnection } from 'typeorm';
+import { Connection, createConnection } from 'typeorm';
```
polymath-eric, line 1: I didn't realize we had `typeorm` in this project.

My preference would be to use sequlize, since that is already a dependency of SQ and I don't want to risk bad interactions between ORMs (they might be OK together, but might not).

If you want to use typeorm, then you can use its migration features to change tables at least?

docker-compose.yml
```diff
@@ -30,6 +30,9 @@ services:
...
+      CURRENT_SQ_VERSION: 5.4.3
```
polymath-eric, line 35:
```suggestion
```

If we can't use an ORM migration tool, then we should be able to read this somehow.

scripts/run-sql.ts
```diff
@@ -62,7 +93,9 @@ const main = async () => {
   console.log('Applied initial SQL');
 
   await postgres.query(migrationQueries().join('\n'));
```
polymath-eric, line 95:
```suggestion
  const migrations = migrationQueries().join('\n');
  await postgres.query(migrations)
```

scripts/run-sql.ts
```diff
@@ -27,6 +27,37 @@ const retry = async <T>(
...
+      .map(file => file.substring(0, file.indexOf('.sql')))
+      .filter(
+        file =>
+          getSQVersion(file) > getSQVersion(PREVIOUS_SQ_VERSION) &&
+          getSQVersion(file) <= getSQVersion(CURRENT_SQ_VERSION)
+      );
+
+    const promises = migrationsToRun.map(migration => {
+      console.log(`Running migration - ${migration}`);
+      return postgres.query(readFileSync(`../db/migrations/${migration}.sql`, 'utf-8'));
```
polymath-eric, line 53: I think `../db/migrations/` should be in a `migrationPath` variable.

docker-compose.yml
polymath-eric: I think `subqueries` table might record the version of SQ running, but its empty when I look at polymesh-local table.

If not we should probably mark a table with that information somehow

polymath-eric (changes requested):
I think there is a better interface for this. 

I am thinking it should be a flag `DB_AUTO_MIGRATE`, default to false, and also document how to run the script manually (e.g. pull the repo and run `yarn db:migrations:run` with appropriate config)

polymath-eric (commented):
I left a few ideas, but if you take out 

`CURRENT_SQ_VERSION` rename `PREVIOUS_SQ_VERSION` to include `DB` I'd be OK approving.

I can't think of a case where we'd have a migrations file that is ahead of the version its in.

db/utils.ts
```diff
@@ -0,0 +1,47 @@
+import { Connection, createConnection } from 'typeorm';
+import { env, chdir } from 'process';
+chdir(__dirname);
+
+require('dotenv').config(); // eslint-disable-line @typescript-eslint/no-var-requires
```
polymath-eric, line 5:
```suggestion
import * as dotenv from 'dotenv';

dotenv.config();
```

db/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    onRetry();
+    await sleep(ms);
+  }
+  throw err;
+};
+
+export const getPostgresConnection = async (): Promise<Connection> => {
+  return retry(
+    env.NODE_ENV === 'local' ? 10 : 1,
```
polymath-eric, line 31:
```suggestion
    const maxAttempts = env.NODE_ENV === 'local' ? 10 : 1,
    
    retry(maxAttempts,...
```

README.md
```diff
@@ -34,3 +34,4 @@ More advanced options are:
 
 - `NO_NATIVE_GRAPHQL_DATA` — this will set the event handler to only record what is necessary for tooling gql. This allows a stable indexer to be ran as native GraphQL handlers get developed
 - `MAX_OLD_SPACE_SIZE` — this will be passed onto the node process as `--max-old-space-size` flag. The recommendation is for this to be ~75% of available RAM. Defaults to 1536, a setting appropriate for 2GB.
+- `CURRENT_SQ_VERSION` & `PREVIOUS_SQ_VERSION` - this will automatically run the required migration files to migrate your database, if any, from previously run subquery version to current one. This can be used to avoid resync and incorporate schema changes required for current version.
```
polymath-eric, line 37: Are these needed still?

prashantasdeveloper: No I'll remove this

VictorVicente (approved):

polymath-eric (approved):

PR #98, opened 2022-11-11
chore: 🤖 Update subql-node to `1.13.3`

prashantasdeveloper:
### Description

Updates subql-node to latest `1.13.3`.

### Breaking Changes

NA

### JIRA Link

DA-401

### Checklist


src/mappings/serializeLikeHarvester.ts
```diff
@@ -76,6 +76,8 @@ export const serializeLikeHarvester = (
     return removeNullChars(item.toString());
   } else if (type === 'Ticker' || type === 'PolymeshPrimitivesTicker') {
     return serializeTicker(item);
+  } else if (rawType.startsWith('[u8;') && rawType.endsWith(']')) {
+    return item.toHex();
```
prashantasdeveloper, line 80: Types such as IdentityId, CddId etc weren't properly serialized with latest changes, so had to add this in.

docker-compose.yml
```diff
@@ -53,6 +57,8 @@ services:
       - --db-schema=public # Once tooling-gql is deprecated, we can change the schema name if required
       # - --force-clean # To remove all tables and schema before starting the app
       - --timestamp-field # Enables created_at and updated_at in schema
+      - --disable-historical=true
+      # - --workers=4 # Number of available CPU cores strictly limits the usage of worker threads. Read more here - https://academy.subquery.network/run_publish/references.html#w-workers
```
prashantasdeveloper, line 61: I'm still not confident with workers so have kept it commented (Can remove as well). 

I was testing some blocks in my local and made it fail at a particular block. SQ kept retrying 3-4 times and altogether skipped the block.

src/chainTypes/harvesterTypes.json
```diff
@@ -96,5 +96,6 @@
   "AuthorityList": "Vec<NextAuthority>",
   "ElectionCompute": {
     "_enum": ["OnChain", "Signed", "Authority"]
-  }
+  },
+  "DispatchErrorModule": "DispatchErrorModuleU8"
```
prashantasdeveloper, line 100: Adding this made it pass the block, which kept on failing the previous time. ([#6918](https://polymesh-testnet.subscan.io/block/0x633b8e81e111b8e9353c99adef52c2def3224b69a2e47280ba34c15553b1cb4e?tab=extrinsic) in testnet)

prashantasdeveloper (commented):
I'm going to start up a separate testnet sync on dev server with this change to make sure all events are processed correctly.

docker-compose.yml
polymath-eric: What do you mean skipped the block? Like it didn't process it and moved onto the next?

That seems bad since it if had createAsset call then it would crash whenever something happened with the Asset

metadata.json
```diff
@@ -1 +1,5 @@
...7472616374735772617070657214484d657461646174614f6654656d706c6174650101062c436f6465486173683c543e7854656d706c6174654d657461646174613c42616c616e63654f663c543e3e0084000324756e646566696e656400000000000000000000000000000000000000000004e02053746f726520746865206d6574612064657461696c73206f662074686520736d61727420657874656e73696f6e2074656d706c6174652e3054656d706c617465496e666f0101062c436f6465486173683c543e7454656d706c61746544657461696c733c42616c616e63654f663c543e3e00c40000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004ec2053746f7265207468652064657461696c73206f66207468652074656d706c617465202845782d206f776e65722c2066726f7a656e20657463292e34457874656e73696f6e496e666f01010630543a3a4163636f756e74496484457874656e73696f6e417474726962757465733c42616c616e63654f663c543e3e0050000000000000000000000000000000000000000004882044657461696c73206f6620657874656e73696f6e2067657420757064617465642e38457874656e73696f6e4e6f6e636501000c75363420000000000000000008d4204e6f6e636520666f722074686520736d61727420657874656e73696f6e206163636f756e742069642067656e65726174696f6e2e5501205573696e67206578706c69636974206e6f6e636520617320696e206261746368207472616e73616374696f6e20616363
```
polymath-eric, line 1: is this supposed to be git ignored now?

As it is we'll always have this version included

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "polymesh-subquery",
-  "version": "5.5.0",
```
polymath-eric, line 3: Maybe a master -> alpha is needed?

package.json
```diff
@@ -85,9 +85,12 @@
     "prettier-eslint-cli": "5.0.1",
     "semantic-release": "^19.0.5",
     "ts-jest": "^27.0.4",
-    "ts-node": "^8.10.2",
+    "ts-node": "^10.9.1",
     "typescript": "^4.1.3"
   },
+  "resolutions": {
+    "ipfs-unixfs": "6.0.6"
```
polymath-eric, line 92: Why does this need to be pinned?

spec_diffs/2021-2023
```diff
@@ -40,6 +40,7 @@
 -     "_enum": Object {
 -       "CINS": "[u8; 9]",
 -       "CUSIP": "[u8; 9]",
+-       "FIGI": "[u8; 12]",
```
polymath-eric, line 43: Isn't this an old file. I wasn't expecting this one to get changed

polymath-eric (commented):
some comments, overall looks good

.gitignore
```diff
@@ -55,3 +55,5 @@ coverage
 
 project.yaml
 tsconfig.tsbuildinfo
+
+metadata.json
```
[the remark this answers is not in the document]
prashantasdeveloper: Yes this was not needed. It was added by Diego long back but was no longer needed.

metadata.json
prashantasdeveloper: We don't need that. If we start using a lot of storage functions or rpcs in our mapping handlers, then to type them strongly we can add all the augment types as we do in SDK. But for now can be removed

package.json
prashantasdeveloper: Not needed. My idea with package.json version was to us for migrations and whenever we think we need some migrations then only its necessary to bump this up to next possible release version. I did it here just to make it sync with latest version.

package.json
prashantasdeveloper: When I created a separate starter project with subql, it had it in there so I added it here as well

spec_diffs/2021-2023
prashantasdeveloper: Diff files are a bit missed up, I have fixed the sequence and generated fresh diffs again

package.json
polymath-eric: Do we need overrides and pnpm like the SDK?

I think just yarn support is fine, although not needing it would be best

docker-compose.yml
prashantasdeveloper: Moved on to the next one

VictorVicente (approved):

PR #100, opened 2022-11-18
feat: 🎸 add generate legacy stat script

polymath-eric:
### Description

We should do a more maintainable way, but this lets us get SQ unstuck on testnet at least. The proposed upstream solution fits this use case perfectly.

### Breaking Changes

None

### JIRA Link

[DA-464](https://polymesh.atlassian.net/browse/DA-464)

### Checklist


polymath-eric:
List obtained from:

```
curl 'http://dev.server:3007/' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://dev.server:3007/' --data-binary '{"query":"query issued {\n  events(filter: {blockId: {greaterThanOrEqualTo: \"1978253\"},eventId: {equalTo: Issued}}) {\n    nodes{\n      id\n      eventArg0\n      eventArg1\n      eventArg2\n      eventArg3\n    }\n  }\n}\n"}' --compressed | jq '.data.events.nodes | map(.eventArg1) | unique'
```

block number was found with:
```
query blocks {
  blocks (first: 1, filter: {specVersionId: {greaterThanOrEqualTo: 3010}}){
    nodes{
      id
      specVersionId
    }
  }
}
```

polymath-eric:
I changed this up to use the built in SubQuery handler.

We can't execute raw SQL so I updated it to use a JS object and the chain name to run it through TypeORM to make the stats.

This needs to be tested first, but it should work and be extensible for future migrations.

db/codeUpdated/mainnet/legacyStats.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO stat_types (id, asset_id, op_type, created_block_id, updated_block_id, created_at, updated_at ) VALUES ('20220824/Count', '20220824', 'Count', '1', '1', NOW(), NOW());
```
polymath-eric, line 1: this should probably be removed, but for now this is the patch to fix current versions

project.template.yaml
```diff
@@ -29,3 +29,8 @@ dataSources:
           kind: substrate/CallHandler
         - handler: ${EVENT_HANDLER}
           kind: substrate/EventHandler
+        - handler: handleChainMigration
```
polymath-eric, line 32: This probably should be a regular event handler instead of changing this file

src/mappings/chainMigrations/codeUpdated.ts
```diff
@@ -0,0 +1,31 @@
...
+        assetId,
+        opType: StatOpTypeEnum.Count,
+        claimType: null,
+        claimIssuerId: null,
+        createdBlockId: blockId,
+        updatedBlockId: blockId,
+      }).save()
+    );
+
+    await Promise.all(newStats);
```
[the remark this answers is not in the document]
prashantasdeveloper: Ideally, we should read in the old data and create new data corresponding to that. Added more around this [here](https://polymesh.slack.com/archives/C044Z31TE56/p1669025798507959)

prashantasdeveloper (changes requested):
I have added the queries that should be used to migrate the data. Those are generic and don't require peeking in database for finding out tickers and all. Take a look [here](https://polymesh.slack.com/archives/C044Z31TE56/p1669026519697249?thread_ts=1669025798.507959&cid=C044Z31TE56)

src/mappings/chainMigrations/legacyStats.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Assets issued before v5.0 had an implicitly made Count statistics
+ *
+ * This is a list for testnet and mainnet obtained with:
+ *
+ * ```sh
+ * curl 'http://localhost:3007/' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://localhost:3007' --data-binary '{"query":"query issued {\n  events(filter: {blockId: {lessThan: \"4397818\"},eventId: {equalTo: Issued}}) {\n    nodes{\n      id\n      eventArg0\n      eventArg1\n      eventArg2\n      eventArg3\n    }\n  }\n}\n"}' --compressed | jq '.data.events.nodes | map(.eventArg1) | unique' | tee /tmp/asset_list.txt
```
prashantasdeveloper, line 7: This query will not give correct data as the blockId column is string and querying on that didn't give correct data. For any block starting with number greater than 4 like 500 or 60001 or 99999 etc, those blocks will get skipped by this.

polymath-eric: Thanks. I was wondering what was wrong with the query.

Any ideas to stop other users from making them same mistake?

prashantasdeveloper: I'm not sure. There should a way around the graphql to first filter out blocks and then fetch in the events but I'm not sure. I'm not that good at clubbing graphql queries

src/mappings/entities/mapTransferManager.ts
```diff
@@ -19,25 +28,58 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+      id,
+      assetId: ticker,
+      type,
+      value,
+      exemptedEntities: [],
+      createdBlockId: blockId,
+      updatedBlockId: blockId,
+    }).save(),
+    TransferCompliance.create({
+      id,
```
prashantasdeveloper, line 48: TransferCompliance id in these cases should be mapped as ticker/opType. Also, StatType creation promise should be pushed before creating TransferCompliance

src/mappings/entities/mapTransferManager.ts
```diff
@@ -19,25 +28,58 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+      assetId: ticker,
+      type: complianceType,
+      statTypeId: `${ticker}/${type}`,
```
prashantasdeveloper, line 51: this should be ticker/opType

src/mappings/entities/mapTransferManager.ts
```diff
@@ -19,25 +28,58 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
 
   const ticker = serializeTicker(rawTicker);
   const { type, value } = getTransferManagerValue(rawManager);
+  const id = `${ticker}/${type}/${value}`;
+  const complianceType =
+    type === TransferRestrictionTypeEnum.Percentage
+      ? TransferComplianceTypeEnum.MaxInvestorOwnership
+      : TransferComplianceTypeEnum.MaxInvestorCount;
```
prashantasdeveloper, line 35:
```suggestion
  let complianceType = TransferComplianceTypeEnum.MaxInvestorCount;
  let opType = StatOpTypeEnum.Count;
  if (type === TransferRestrictionTypeEnum.Percentage) {
     complianceType = TransferComplianceTypeEnum.MaxInvestorOwnership; 
     opType = StatOpTypeEnum.Balance;
  }
```

src/mappings/entities/mapTransferManager.ts
```diff
@@ -19,25 +28,58 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+      exemptedEntities: [],
+      createdBlockId: blockId,
+      updatedBlockId: blockId,
+    }).save(),
+    TransferCompliance.create({
+      id,
+      assetId: ticker,
+      type: complianceType,
+      statTypeId: `${ticker}/${type}`,
+      value: null,
```
prashantasdeveloper, line 52: value should be the value for transfer manager

src/mappings/entities/mapTransferManager.ts
```diff
@@ -19,25 +28,58 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
 
   const ticker = serializeTicker(rawTicker);
   const { type, value } = getTransferManagerValue(rawManager);
+  const id = `${ticker}/${type}/${value}`;
 
-  await TransferManager.remove(`${ticker}/${type}/${value}`);
+  await Promise.all([TransferManager.remove(id), TransferCompliance.remove(id)]);
 };
 
 const handleExemptionsAdded = async (blockId: string, params: Codec[]): Promise<void> => {
```
prashantasdeveloper, line 85: You forgot to handle the exemptions case here where we need to add to TransferComplianceExemptions

src/mappings/mappingHandlers.ts
```diff
@@ -167,6 +168,7 @@ export async function handleEvent(event: SubstrateEvent): Promise<void> {
       moduleId: moduleId as ModuleIdEnum,
       params: args,
       event,
+      specVersion,
```
prashantasdeveloper, line 171: nit -> Was there a need to pass this as handlerArgs? we could have got it from event itself (event.block.specVersion)? I don't have problem if you leave it as it as

src/mappings/entities/mapAsset.ts
```diff
@@ -240,10 +243,29 @@ const handleIssued = async (
...
+    promises.push(
+      StatType.create({
+        id: `${ticker}/Count`,
+        opType: StatOpTypeEnum.Count,
+        assetId: ticker,
+        claimType: null,
+        claimIssuerId: null,
+        createdBlockId: blockId,
+        updatedBlockId: blockId,
+      }).save()
```
prashantasdeveloper, line 257: You should check if stat type is already present or not before creating it

prashantasdeveloper (changes requested):

src/mappings/mappingHandlers.ts
polymath-eric: I didn't realize I had a connection to the block. Will change

src/mappings/entities/mapTransferManager.ts
polymath-eric: I moved it up since thats the intention, but if there is a race condition I am not sure dispatching the promise earlier is enough to guarantee it gets reflected in the DB first.

Hopefully it isn't an issue.

db/migrations/8.4.1.sql
```diff
@@ -0,0 +1,65 @@
+with asset_data as (
```
polymath-eric, line 1: @prashantasdeveloper I think you took 8.4.2 in master branch.

I want to merge this into alpha so we can test resync. Does this need to get 8.4.3 now, or will filling in the "hole" be OK?

src/mappings/entities/mapTransferManager.ts
```diff
@@ -74,8 +164,16 @@ const handleExemptionsRemoved = async (blockId: string, params: Codec[]) => {
...
     transferManager.updatedBlockId = blockId;
 
-    await transferManager.save();
+    promises.push(transferManager.save());
   }
+
+  const transferComplianceExemptions = await TransferComplianceExemption.getByAssetId(ticker);
+
+  transferComplianceExemptions
+    .filter(({ exemptedEntityId }) => parsedExemptions.includes(exemptedEntityId))
```
polymath-eric, line 173: Pretty sure this needs to filter by type as well

src/mappings/entities/mapTransferManager.ts
```diff
@@ -29,12 +30,35 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+  let complianceType = TransferComplianceTypeEnum.MaxInvestorCount;
+  let opType = StatOpTypeEnum.Count;
+  if (type === TransferRestrictionTypeEnum.Percentage) {
+    complianceType = TransferComplianceTypeEnum.MaxInvestorOwnership;
+    opType = StatOpTypeEnum.Balance;
+  }
+
+  const promises = [];
+
+  if (complianceType === TransferComplianceTypeEnum.MaxInvestorOwnership) {
```
prashantasdeveloper, line 43:
```suggestion
  const promises = [];
  if (type === TransferRestrictionTypeEnum.Percentage) {
    complianceType = TransferComplianceTypeEnum.MaxInvestorOwnership;
    opType = StatOpTypeEnum.Balance;
```

src/mappings/entities/mapTransferManager.ts
```diff
@@ -29,12 +30,35 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+    const statId = `${ticker}/Balance`;
```
prashantasdeveloper, line 44:
```suggestion
    const statId = `${ticker}/${StatOpTypeEnum.Balance}`;
```

src/mappings/entities/mapTransferManager.ts
```diff
@@ -91,19 +104,54 @@ const handleExemptionsAdded = async (blockId: string, params: Codec[]): Promise<
...
+            ...exemptKey,
+            exemptedEntityId: entity,
+            createdBlockId: blockId,
+            updatedBlockId: blockId,
+          });
+        }
+        return exemption.save();
+      };
+      promises.push(upsert());
+    });
```
prashantasdeveloper, line 145: instead of re-writing this I think we should extract the method in `mapStatistics` file and reuse here

src/mappings/entities/mapTransferManager.ts
```diff
@@ -29,12 +30,35 @@ const handleTransferManagerAdded = async (blockId: string, params: Codec[]) => {
...
+      promises.push(
+        StatType.create({
+          id: statId,
+          opType: StatOpTypeEnum.Balance,
+          assetId: ticker,
+          claimType: null,
+          claimIssuerId: null,
+          createdBlockId: blockId,
+          updatedBlockId: blockId,
+        }).save()
```
prashantasdeveloper, line 56: We should extract this code in statistics handler and reuse here

prashantasdeveloper (changes requested):
Overall LGTM, but left some comments around refactoring the code. My idea is to limit handlers to define the related entities and wherever need be, other handlers can call those methods. That way, we limit the changes to an entity to a single file

db/migrations/8.4.1.sql
prashantasdeveloper: This would require 8.4.3

src/mappings/entities/mapStatistics.ts
```diff
@@ -256,11 +268,143 @@ const handleExemptionsRemoved = async (params: Codec[]) => {
...
+        id: statId,
+        opType: StatOpTypeEnum.Balance,
+        assetId: ticker,
+        claimType: null,
+        claimIssuerId: null,
+        createdBlockId: blockId,
+        updatedBlockId: blockId,
+      }).save();
+    }
+  }
```
prashantasdeveloper, line 291: You could refactor this as 
```
const createStatType = async (
  { assetId, opType, claimType, claimIssuerId }: Attributes<StatType>,
  blockId: string
) => {
  const statTypeId = getStatTypeId(assetId, opType, claimType, claimIssuerId);
  let statType = await StatType.get(statTypeId);
  if (statType) {
    statType.updatedBlockId = blockId;
  } else {
    statType = StatType.create({
      id: statTypeId,
      assetId,
      opType,
      claimType,
      claimIssuerId,
      createdBlockId: blockId,
      updatedBlockId: blockId,
    });
  }
  return statType.save();
};
```

And call this method wherever StatType.create is invoked. Similarly other code refactoring can be done so that, we have single point from where StatType.create can be called

prashantasdeveloper (changes requested):

prashantasdeveloper (approved):

PR #111, opened 2023-01-05
fix: 🐛 Query fix in legacy stat script

prashantasdeveloper:
### Description

There was an error in the legacy stat query which I fixed and pushed directly into master by mistake ([fix: 🐛 Query fix in legacy stat script](https://github.com/PolymeshAssociation/polymesh-subquery/commit/75e7bb3a99cdf12d8fde3a222c2924b052457d83)). Adding the change with fix commit to bump up the version 

### Breaking Changes

NA

### JIRA Link

NA

### Checklist


prashantasdeveloper:
Merging it quickly to have version right now as Marcin is trying to update the testnet with this.

PR #112, opened 2023-01-09
feat: 🎸 Add AssetTransaction entity

prashantasdeveloper:
### Description

Adds AssetTransaction entity

### Breaking Changes

NA

### JIRA Link

DA-195

### Checklist


src/mappings/entities/mapAsset.ts
```diff
@@ -286,32 +302,73 @@ const handleAssetOwnershipTransferred = async (blockId: string, params: Codec[])
...
-    getAssetHolder(ticker, fromDid, blockId),
-    getAssetHolder(ticker, toDid, blockId),
-  ]);
+  const promises = [];
 
-  fromHolder.amount = fromHolder.amount - transferAmount;
-  fromHolder.updatedBlockId = blockId;
-  toHolder.amount = toHolder.amount + transferAmount;
-  toHolder.updatedBlockId = blockId;
+  if (fromDid === '0x00'.padEnd(66, '0')) {
```
polymath-eric, line 317: Can we assign `'0x00'.padEnd(66, '0')` to a const and use that? Something like `emptyDid`

polymath-eric (commented):
Seems like the test case got stuck btw

prashantasdeveloper:
Created a doc listing all the events - https://polymesh.atlassian.net/wiki/spaces/EN/pages/4114710529/Draft+Asset+Transactions. Based on the changes there, some minor changes can occur in the PR

polymath-eric (commented):
LGTM, left a comment about improving our GraphQL documentation, which I think could clarify the "null" portfolio case a lot for people.

Might be worthwhile comparing Asset amounts with the chain to double check we didn't miss any event. More work would be to add that explicitly into SQ, but we should be able to write a "SELECT SUM(amount) FROM asset_transfers GROUP BY asset..." (not sure about making "from" a negative though)

src/mappings/entities/mapAsset.ts
```diff
@@ -286,32 +303,76 @@ const handleAssetOwnershipTransferred = async (blockId: string, params: Codec[])
...
-  ]);
+  const promises = [];
 
-  fromHolder.amount = fromHolder.amount - transferAmount;
-  fromHolder.updatedBlockId = blockId;
-  toHolder.amount = toHolder.amount + transferAmount;
-  toHolder.updatedBlockId = blockId;
+  if (fromDid === emptyDid) {
+    fromPortfolioId = null;
+    return; // We ignore the transfer case when Asset tokens are issued
```
polymath-eric, line 320: why don't we track issue transaction? are they being added in another place?

prashantasdeveloper: Yes, the are tracked in AssetCreated

VictorVicente (approved):

PR #113, opened 2023-01-10
feat: 🎸 Add PolyxTransaction entity

prashantasdeveloper:
### Description

Adds a new entity to track all the POLYX transactions

### Breaking Changes

NA

### JIRA Link

DA-195

### Checklist


polymath-eric (commented):
LGTM

Will look again once we have all events

src/mappings/entities/mapPolyxTransaction.ts
```diff
@@ -39,11 +41,38 @@ const handleTreasuryReimbursement = async (args: HandlerArgs): Promise<void> =>
 
   const did = getTextValue(rawIdentity);
   const identity = await Identity.get(did);
+
+  // treasury reimbursement is only 80% of the actual amount deducted
+  const reimbursementBalance = new BigNumber(getTextValue(rawBalance) || 0);
```
polymath-eric, line 46: I don't get this. Why wouldn't it be 100%, and why wouldn't the actual amount not be on the event.

polymath-eric (commented):
I don't get the 80% part. Multiplying by a float value seems risky here from a numbers lining up 100% point of view. I trust BigInt wouldn't cause us issues, but it seems kind of arbitrary to me as well.

Does the chain expose this ratio in an RPC somewhere? Is it a constant for all blocks always, or can it be adjusted?

db/migrations/9.4.05.sql
```diff
@@ -0,0 +1,44 @@
...
+        create type public_enum_5df0f1d22c AS ENUM ('Free', 'Reserved', 'Bonded', 'Unbonded', 'Locked');
+    END IF;
+END
+$$;
+
+create table if not exists polyx_transactions
+(
+  id text not null PRIMARY KEY,
+  identity_id text,
+  "address" text,
```
polymath-eric, line 14:
```suggestion
  address text,
```

db/migrations/9.4.05.sql
```diff
@@ -0,0 +1,44 @@
...
+  to_id text,
+  to_address text,
+  amount numeric not null,
+  "type" public_enum_5df0f1d22c not null,
```
polymath-eric, line 18: are these quotes intended?

scripts/run-sql.ts
```diff
@@ -22,7 +22,7 @@ const main = async (): Promise<void> => {
     await postgres.query(readFileSync('../db/compat.sql', 'utf-8'));
     console.log('Applied initial SQL');
 
-    await postgres.query(migrationQueries().join('\n'));
+    await postgres.query((await migrationQueries()).join('\n'));
```
polymath-eric, line 25: (style) no inline await

src/mappings/entities/mapPolyxTransaction.ts
```diff
@@ -0,0 +1,532 @@
...
+     * and skip adding a separate `TreasuryDisbursement` entry
+     */
+    if (transferPolyxTransaction && amount === transferPolyxTransaction.amount) {
+      // this is the case where treasury reimbursement is showing that 80% of protocol fee charged
+      // We ignore this case to insert in PolyxTransaction
+      Object.assign(transferPolyxTransaction, {
+        eventId: EventIdEnum.TreasuryDisbursement,
+      });
+      await transferPolyxTransaction.save();
+      return;
```
polymath-eric, line 150: (style, if the linter is OK with it)
```suggestion
```

src/mappings/entities/mapPolyxTransaction.ts
prashantasdeveloper: it was like this before 5.4.1. Not sure why, but they changed it since 5.4.1

VictorVicente (approved):

PR #117, opened 2023-01-31
feat: 🎸 handlePrimaryKeyUpdated & handleSecondaryKeyLeftIdentiy

sansan:
### Description

handle handlePrimaryKeyUpdated and handleSecondaryKeyLeftIdentity
 
### JIRA Link

https://polymesh.atlassian.net/browse/DA-471

### Checklist


src/mappings/entities/mapIdentities.ts
```diff
@@ -361,3 +369,39 @@ const handleSecondaryKeysAdded = async (
...
+
+  const account = rawAccount.toString() as MeshAccount;
+
+  let address;
+  if (typeof account === 'string') {
+    // for chain version >= 5.0.0
+    address = account;
+  } else {
+    // for chain version < 5.0.0
+    ({ account: address } = account);
```
polymath-eric, line 403: `account` is only made from a `.toString` call, so any destructure will be a run time error.

I think we do the `.toString`, then feed that to `JSON.parse` for this to work

src/mappings/entities/mapIdentities.ts
```diff
@@ -361,3 +369,39 @@ const handleSecondaryKeysAdded = async (
...
+  await identity.save();
+};
+
+const handleSecondaryKeyLeftIdentity = async (params: Codec[]): Promise<void> => {
+  const [, rawAccount] = params;
+
+  const account = rawAccount.toString() as MeshAccount;
+
+  let address;
+  if (typeof account === 'string') {
```
polymath-eric, line 398: This is the case 100% of the time

src/mappings/entities/mapIdentities.ts
```diff
@@ -50,6 +50,14 @@ export async function mapIdentities({
     if (eventId === EventIdEnum.SecondaryKeyPermissionsUpdated) {
       await handleSecondaryKeysPermissionsUpdated(blockId, params);
     }
+
+    if (eventId === EventIdEnum.PrimaryKeyUpdated) {
+      await handlePrimaryKeyUpdated(blockId, eventId, params);
+    }
+
+    if (eventId === EventIdEnum.SecondaryKeyLeftIdentity) {
```
polymath-eric, line 58: Sonar cloud complains because these are nested which doubles their complexity value.

If you switch the top conditional to be an early return, the complexity will drop well under the threshold.

I prefer to change the code since the smells might need to get reignored in the future, although this is OK to ignore if you prefer.

src/mappings/entities/mapIdentities.ts
```diff
@@ -361,3 +371,39 @@ const handleSecondaryKeysAdded = async (
...
+  let address;
+  if (typeof account === 'string') {
+    // for chain version >= 5.0.0
+    address = account;
+  } else {
+    // for chain version < 5.0.0
+    ({ account: address } = account);
+  }
+
+  await Promise.all([[Account.remove(address), Permissions.remove(address)]]);
```
prashantasdeveloper, line 408: In this scenario, I would only remove the mapping of DID from Account instead of removing the entire account.

prashantasdeveloper (changes requested):
This would also require a migration to populate and fix the existing data

src/mappings/entities/mapIdentities.ts
prashantasdeveloper: What do you think @polymath-eric ?

src/mappings/entities/mapIdentities.ts
```diff
@@ -361,3 +371,39 @@ const handleSecondaryKeysAdded = async (
...
+    updatedBlockId: blockId,
+    eventId,
+  });
+
+  await identity.save();
+};
+
+const handleSecondaryKeyLeftIdentity = async (params: Codec[]): Promise<void> => {
+  const [, rawAccount] = params;
+
```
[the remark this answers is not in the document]
prashantasdeveloper: Right now when secondary keys are removed, we remove both Account and Permission entry, which now that I think of I'm not sure we should be removing altogether or just updating in a way to showcase that the Account is unlinked?

db/migrations/9.2.06.sql
```diff
@@ -0,0 +1,62 @@
...
+    event_arg_0 as did,
+    CASE
+      WHEN event_id = 'SecondaryKeysFrozen' THEN true
+      WHEN event_id = 'SecondaryKeysUnfrozen' THEN false
+      ELSE null
+    END as is_frozen,
+    created_at as updated_at
+  FROM events
+  WHERE event_id IN ('SecondaryKeysFrozen', 'SecondaryKeysUnfrozen')
+  ORDER BY created_at DESC
```
prashantasdeveloper, line 17: Why do we need to updated the frozen field? Has something changed ?

db/migrations/9.2.06.sql
```diff
@@ -0,0 +1,62 @@
...
+SET
+  secondary_keys_frozen = event_data.is_frozen,
+  updated_block_id = event_data.block_id,
+  updated_at = event_data.updated_at
+FROM event_data
+WHERE identities.id = event_data.did;
+
+-- handle secondary key left identity 
+
+WITH event_data AS (
```
prashantasdeveloper, line 29: I left some comments below, based on the discussion there we may require to remove the account entry as well or update something here

db/migrations/9.2.06.sql
```diff
@@ -0,0 +1,62 @@
...
+    created_at as updated_at,
+    event_id as event_id
+  FROM events
+  WHERE event_id = 'PrimaryKeyUpdated'
+  ORDER BY created_at DESC
+)
+UPDATE identities
+SET
+  primary_account = event_data.new_key,
+  event_id = event_data.event_id,
```
prashantasdeveloper, line 56:
```suggestion

```

db/migrations/9.2.06.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- handle primary key updated
+
+WITH event_data AS (
+  SELECT
+    event_arg_0 as did,
+    event_arg_2 as new_key,
+    block_id as updated_block_id,
+    created_at as updated_at,
+    event_id as event_id
```
prashantasdeveloper, line 48:
```suggestion
```

prashantasdeveloper (changes requested):
I left a few comments around the overall process as well. Maybe we need to discuss this with the team before proceeding

src/mappings/entities/mapIdentities.ts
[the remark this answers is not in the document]
polymath-eric: I don't think we should use a bool to track it. Its too hard to document to users imo.

Could we have an `AccountIdentity` history? Instead of deleting the record, move it to a history table instead.

I think the `drop not null` constraint makes sense, so the "natural" queries reflect the chain state.

db/migrations/9.4.04.sql
```diff
@@ -0,0 +1,247 @@
...
+         FROM information_schema.columns 
+         WHERE table_schema = 'public'
+               AND table_name = 'accounts'
+               AND column_name = 'identity_id'
+               AND is_nullable = 'NO' )
+      THEN
+         ALTER TABLE public.accounts ALTER COLUMN identity_id DROP NOT NULL;
+      END IF;
+   END IF;
+END $$;
```
prashantasdeveloper, line 22: If you directly add in the alter query does that fail in some case? I think altering column to drop not null would not throw an error if the column was already nullable in the first place. Worth double checking

db/migrations/9.4.04.sql
```diff
@@ -0,0 +1,247 @@
...
+
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
prashantasdeveloper, line 24: We should avoid adding any new extensions. Idea is to have a db state which would be same if a user starts syncing from scratch without running migrations

db/migrations/9.4.04.sql
```diff
@@ -0,0 +1,247 @@
...
+
+CREATE TABLE IF NOT EXISTS account_history (
+    id text NOT NULL DEFAULT uuid_generate_v4(),
+    account_id text,
+    identity_id text,
+    event_id public_enum_8f5a39c8ee,
+    permissions_id text,
+    datetime timestamp(6),
+    updated_block_id text
+);
```
prashantasdeveloper, line 34: We always added `created_block_id`, `updated_block_id` to all the entities, and we don't use uuid generators for filling in the PK sequence (because SQ doesn't follow this). Idea is to have the same table structure as the SQ does. 

Also, you forgot to add constraints and indexes

What I generally do is run an SQ instance with the new table added and then check the table structure using `\d+ tablename` . This gives me the exact idea on which constraints I need to add, which indexes etc. etc

db/migrations/9.4.04.sql
```diff
@@ -0,0 +1,247 @@
...
+  WHERE event_id = 'PrimaryKeyUpdated'
+  ORDER BY created_at ASC
+), account_data AS (
+  SELECT
+    accounts.id,
+    accounts.permissions_id
+  FROM accounts
+  JOIN event_data ON accounts.identity_id = event_data.did
+)
+INSERT INTO account_history (account_id, identity_id, event_id, permissions_id, datetime, updated_block_id)
```
prashantasdeveloper, line 57: By default SQ adds `created_at` and `updated_at` column to each table in SQ, those details are missing in the create table query as well and in the insert query as well

prashantasdeveloper (changes requested):
Left more comments. Idea with migration is - 

- to make sure that even if I execute migrations multiple times I should not get repeated data
- to have a SQ state that would match a SQ state with full sync without running migration. So keep in mind to have exact same structure as the SQ does

db/migrations/9.4.04.sql
```diff
@@ -221,14 +217,18 @@ WITH event_data AS (
...
+INSERT INTO account_history (id, account_id, identity_id, event_id, permissions_id, datetime, created_block_id, updated_block_id, created_at, updated_at)
 SELECT
+  event_data.id,
   account_data.id,
   account_data.identity_id,
   'SecondaryKeyLeftIdentity',
   account_data.permissions_id,
-  event_data.updated_at,
-  event_data.updated_block_id
+  event_data.created_at,
```
prashantasdeveloper, line 227: This should be block datetime

prashantasdeveloper (changes requested):

polymath-eric (approved):

prashantasdeveloper (changes requested):
Two nitty comments, other than that LGTM

prashantasdeveloper (approved):

prashantasdeveloper:
This needs bumping the migration number

db/migrations/3_identity_history.sql
```diff
@@ -0,0 +1,262 @@
...
+  event_data.block_id,
+  event_data.block_id,
+  event_data.created_at,
+  event_data.updated_at
+FROM account_data, event_data;
+
+-- select old account permissions and copy them to new account
+WITH event_data AS (
+  SELECT
+    event_arg_0 as did,
```
prashantasdeveloper, line 62: I added a way to get rid of these by using mappingHandlers. Can you check @sansan ?

db/migrations/3_identity_history.sql
```diff
@@ -0,0 +1,262 @@
+ALTER TABLE public.accounts ALTER COLUMN identity_id DROP NOT NULL;
+
+CREATE TABLE IF NOT EXISTS account_history (
+    id text NOT NULL PRIMARY KEY,
+    account_id text NOT NULL REFERENCES accounts(id) ON UPDATE CASCADE,
+    identity_id text REFERENCES identities(id) ON UPDATE CASCADE,
```
prashantasdeveloper, line 6: This should be migration 4 now

prashantasdeveloper (changes requested):

prashantasdeveloper (approved):

PR #118, opened 2023-02-07
fix: 🐛 Update `Sto` entity to save data correctly

prashantasdeveloper:
### Description

Sto entity was overriding data for various offerings based on the ID. For example, if there were two Sto's created lets say for TICKER1 with id 0 and TICKER2 with id 0 in that mentioned order, TICKER1 with 0 was overridden in database with TICKER2 with 0. 

To fix it we have added a new attribute `stoId` to the `Sto` entity and the Sto handler now correctly generates unique ID for Sto as combination of offering Asset and Sto ID fixing the problem.

### Breaking Changes

🧨 Sto ID for an offering Asset is now stored in `stoId` attribute instead of `id` attribute of `Sto`

### JIRA Link

DA-521

### Checklist

### DIFF
--- a/db/migrations/8.5.3.sql
+++ b/db/migrations/8.5.3.sql
@@ -1,0 +1,35 @@
+-- add sto_id column 
+alter table stos
+add column if not exists sto_id integer;
+
+-- create index for sto_id column
+create index if not exists stos_sto_id on stos using btree(sto_id);
+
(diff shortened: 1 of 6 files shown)


polymath-eric (approved):
LGTM

VictorVicente (approved):

PR #121, opened 2023-02-16
feat: 🎸 Update STO entity to track more detailed info

prashantasdeveloper:
### Description

Update STO entity to track more detailed information.
This adds the following new attributes to the existing STO entity
 - `name`, `offeringPortfolio`, `raisingAssetId`, `raisingPortfolio`, `creator`, `start`, `end`, `minimum_investment`, `status`

### Breaking Changes

NA

### JIRA Link

DA-522

### Checklist


db/migrations/9.1.02.sql
```diff
@@ -0,0 +1,251 @@
...
+      )::NUMERIC as minimum_investment,
+    block_id as created_block_id, 
+    block_id as updated_block_id, 
+    created_at as created_at, 
+    updated_at as updated_at
+  from events e
+  where module_id = 'sto' and event_id in ('FundraiserCreated')
+  order by block_id::int
+)
+INSERT INTO "public"."stos" ("id", "sto_id", "name", "venue_id", "offering_asset_id", "offering_portfolio_id",
```
polymath-eric, line 86:
```suggestion
insert into "public"."stos" ("id", "sto_id", "name", "venue_id", "offering_asset_id", "offering_portfolio_id", 
```

seems like it would be more consistent with the smaller case. I think with syntax highlighting all upper case is a bit redundant and more annoying to type. OK to leave.

I think the `::NUMERIC` all upper is good since the case should stand out a bit

src/mappings/entities/mapSto.ts
```diff
@@ -1,22 +1,140 @@
...
+  event: SubstrateEvent,
+  status: StoStatus
+) => {
+  const [, rawStoId] = params;
+  const offeringAssetId = serializeTicker(event.extrinsic.extrinsic.args[0]);
+  const stoId = getNumberValue(rawStoId);
+
+  const sto = await Sto.get(`${offeringAssetId}/${stoId}`);
+
+  if (sto) {
```
polymath-eric, line 43: could this be undefined? shouldn't that be an error?

polymath-eric (approved):

polymath-eric (approved):

PR #146, opened 2023-05-17
fix: 🐛 Negative balance issue fix for asset holders

prashantasdeveloper:
### Description

When tokens for an Asset were issued/redeemed by an external agent, those tokens were getting mapped with the Asset owner instead of the external agent which was messing up the balance for asset holders. This fix maps those tokens with the external agents and corrects the existing data as well.

### Breaking Changes

NA

### JIRA Link

DA-752

### Checklist

### DIFF
--- a/src/mappings/entities/mapAsset.ts
+++ b/src/mappings/entities/mapAsset.ts
@@ -209,8 +209,10 @@ const handleIssued = async (
   params: Codec[],
   event: SubstrateEvent
 ): Promise<void> => {
-  const [, rawTicker, , rawAmount, rawFundingRound, rawTotalFundingAmount] = params;
+  const [, rawTicker, rawBeneficiaryDid, rawAmount, rawFundingRound, rawTotalFundingAmount] =
+    params;
 
+  const issuerDid = getTextValue(rawBeneficiaryDid);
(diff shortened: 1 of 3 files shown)


VictorVicente (approved):

src/mappings/entities/mapAsset.ts
```diff
@@ -244,20 +246,21 @@ const handleIssued = async (
 };
 
 const handleRedeemed = async (blockId: string, params: Codec[]): Promise<void> => {
-  const [, rawTicker, , rawAmount] = params;
+  const [, rawTicker, rawBeneficiaryDid, rawAmount] = params;
```
polymath-eric, line 249:
```suggestion
  const [, rawTicker, rawIssuerDid, rawAmount] = params;
```
(nit) looks more consistent after getting parsed.

polymath-eric: Few other places like this. OK to leave if thats whats called on the event though

polymath-eric (approved):

src/mappings/entities/mapAsset.ts
prashantasdeveloper: The event said it to be beneficiary, so I left it like that

PR #152, opened 2023-05-30
feat: 🎸 Add new migration sequencing

prashantasdeveloper:
### Description

This changes the migration sequencing from package version naming to more generic numeric sequence. 

For the first execution, to be on the safe side, it will execute all the old_migrations present. On later executions, the logic first fetches the last executed migration in db and then based on that fetches all new migrations and executes them. 

All the entries are stored in `Migration` entity that would help us debug without actually looking at SQ node logs

### Breaking Changes

NA

### JIRA Link

DA-762

### Checklist


tests/unit/db.test.ts
```diff
@@ -0,0 +1,13 @@
+import { readdirSync } from 'fs';
+import { getMigrationVersion } from '../../db/schemaMigrations';
+
+test('migration files', () => {
+  const migrations = readdirSync('../db/migrations');
+
+  const invalidFile = migrations.find(file => !getMigrationVersion(file));
+  if (invalidFile) {
+    throw new Error(
+      `Migration file - ${invalidFile} has incorrect pattern. Migration file should always follow the pattern 'V{version}_{description}.sql'. For example 'V1_add_new_entity.sql'`
```
polymath-eric, line 10: is "version" going to be 1, 2, 3? Version doesn't seem like the right word to me, to me I think version is more like the difference between "old" (V1) and "new" (V2), where as each migration shouldn't have multiple versions of it. 

Maybe just a number instead of letter prefix? e.g.
```suggestion
      `Migration file - ${invalidFile} has incorrect pattern. Migration file should always follow the pattern '{id}_{description}.sql'. For example '1_add_new_entity.sql'`
```

db/migrations/V1_migration_entity.sql
```diff
@@ -0,0 +1,10 @@
+create table if not exists migrations (
```
polymath-eric, line 1: I thought SQ itself managed the table creation for us? Shouldn't this table be made because its in the GraphQL schema already and isn't prepopulated with data here?

polymath-eric (commented):
Some nit comments, looks OK, but also feels like we are reinventing a DB migrator too.

Maybe this is worth asking the SQ team about?

Code looks good, but I don't want us to spend time reinventing a generic tool.

tests/unit/db.test.ts
polymath-eric: I know lots of tools will use a dev timestamp to avoid conflicts, but we need to ensure a strict order, so number like you have makes more sense, but they can be hard to coordinate with a big team or long lived PRs

prashantasdeveloper: I tried to match the naming pattern of how flyway does it, but then that has a pattern of V1__example_file.sql, where prefix can be a V, U or R ([link](https://www.red-gate.com/blog/database-devops/flyway-naming-patterns-matter)). Anyways, I was OK with removing the prefix V and sticking to sequential numbering. 

> I know lots of tools will use a dev timestamp to avoid conflicts, but we need to ensure a strict order, so number like you have makes more sense, but they can be hard to coordinate with a big team or long lived PRs

I think that is a good problem to have. I'll be happy to change when there are 15-20 people working on SQ with new features all the times 😄

db/migrations/V1_migration_entity.sql
prashantasdeveloper: There was an instance in the past where the table wasn't created on its own, that's when I started adding new tables as well. I think this happened when we added subquery_versions table. Or maybe that was because of the async execution of the SQLs that we do before running the jest node for SQ. Anyways, I wanted to add a migration to start the sequencing and this does the work, so I'm keeping this for now.

prashantasdeveloper:
> Minor nit, but the JIRA that got linked is going to an old board, even though the number is high enough to only exist on Polymesh board. Is this a auto link setting? Usually I paste my link in manually

This was handled by the Autolink Reference in the Settings - https://github.com/PolymeshAssociation/polymesh-subquery/settings/key_links. I have corrected the link to point it to polymesh one. I did this in other repos as well but somehow this got left out. Thanks for noticing

PR #161, opened 2023-08-04
feat: 🎸 Upgrade subql-node and subql-query to latest versions

prashantasdeveloper:
### Description

Upgrade the SQ images to point to latest versions

### Breaking Changes

NA

### JIRA Link

DA-745

### Checklist


docker-compose.yml
```diff
@@ -65,7 +65,7 @@ services:
       start_period: 30s
 
   graphql-engine:
-    image: onfinality/subql-query:v1.0.0
+    image: onfinality/subql-query:latest
```
polymath-eric, line 68: imo its better to use an explicit version here so it won't break all of a sudden on for someone + we'll be testing with the same versions

docker/sq-Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM onfinality/subql-node:v1.13.3
+FROM onfinality/subql-node:latest
```
polymath-eric, line 1: lets us an explict version here

polymath-eric (commented):
lgtm. I think we should use an explicit version for images since `latest` can grab a breaking version and can be annoying to debug.

polymath-eric (dismissed):

VictorVicente (approved):

polymath-eric (approved):

PR #174, opened 2023-10-23
feat: 🎸 Map NFT events

polymath-eric:
### Description

Add `isNftCollection` to `Asset`. Add `NftHolder` entity. `Identity` now has `heldNfts` field. `Leg`, `PortfolioMovement` and `AssetTransaction` now have `nftIds`. `amount` is now nullable for these entities

### Breaking Changes

BREAKING CHANGE: 🧨  `amount` is now nullable for `Leg`, `PortfolioMovement` and `AssetTransaction`

### JIRA Link

✅ Closes: [DA-905](https://polymesh.atlassian.net/browse/DA-905)

### Checklist

[DA-905]: https://polymesh.atlassian.net/browse/DA-905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1597,7 +1597,7 @@ type PortfolioMovement @entity {
   asset: Asset!
   "the number of fungible tokens transferred. Defined if type == Fungible"
   amount: BigInt
-  "the number of fungible tokens transferred. Defined if type == NonFungible"
+  "the IDs of non-fungible tokens transferred. Defined if type == NonFungible"
   nftIds: [BigInt]
   memo: String
   address: String!
@@ -2099,7 +2099,9 @@ type AssetTransaction @entity {
   fromPortfolio: Portfolio
   "`toPortfolioId` will be null in transactions where the Asset was redeemed"
   toPortfolio: Portfolio
+  "`amount` is the number of fungible tokens involved in this transaction"
   amount: BigInt
+  "`nftIds` are the IDs of the non-fungible tokens involved in this transaction"
   nftIds: [BigInt]
   eventId: EventIdEnum!
   eventIdx: Int!
--- a/schema.graphql
+++ b/schema.graphql
@@ -1348,6 +1348,7 @@ type Identity @entity {
   createdBlock: Block!
   updatedBlock: Block!
   heldAssets: [AssetHolder]! @derivedFrom(field: "identity")
+  heldNfts: [NftHolder]! @derivedFrom(field: "identity")
   portfolios: [Portfolio!]! @derivedFrom(field: "identity")
   children: [ChildIdentity!]! @derivedFrom(field: "parent")
 }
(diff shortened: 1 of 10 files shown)


schema.graphql
```diff
@@ -1578,17 +1579,26 @@ type Portfolio @entity {
...
 type PortfolioMovement @entity {
   id: ID!
   from: Portfolio! @index(unique: false)
   to: Portfolio! @index(unique: false)
+  type: PortfolioMovementTypeEnum!
   asset: Asset!
-  amount: BigInt!
+  "the number of fungible tokens transferred. Defined if type == Fungible"
+  amount: BigInt
+  "the number of fungible tokens transferred. Defined if type == NonFungible"
```
prashantasdeveloper, line 1600:
```suggestion
  "the list of fungible tokens transferred. Defined if type == NonFungible"
```
Is it the count or the actual nft ids?

schema.graphql
```diff
@@ -2045,6 +2068,7 @@ type Asset @entity {
   id: ID! # ticker
   ticker: String! @index(unique: true)
   name: String @index(unique: false)
+  isNftCollection: Boolean!
```
prashantasdeveloper, line 2071: IMO we should add index here `@index(unique: false)` since this would be used a lot in querying the data

src/mappings/entities/common.ts
```diff
@@ -14,3 +14,13 @@ export interface HandlerArgs {
   params: Codec[];
   event: SubstrateEvent;
 }
+
+export const getAsset = async (ticker: string): Promise<Asset> => {
+  const asset = await Asset.getByTicker(ticker);
+
+  if (!asset) {
```
prashantasdeveloper, line 21: IMO this should have been in `mapAsset` itself. I would prefer we input entity types in their respective mappers only (until unless we can't).

schema.graphql
prashantasdeveloper: Also, do you think we should have this type of boolean parameter in `AssetTransaction` as well?

prashantasdeveloper (changes requested):
Left some small comments

schema.graphql
polymath-eric: The actual IDs.

schema.graphql
polymath-eric: I don't think a boolean index helps that much, it doesn't really organize data much since so many things are tied.

I found this SO that has some discussion around it: https://stackoverflow.com/questions/2113181/does-it-make-sense-to-use-an-index-that-will-have-a-low-cardinality

src/mappings/entities/common.ts
polymath-eric: I moved it from `mapAsset`. I think its better to have the functions that are imported into other handlers in a clear common folder, but I agree the shared functions should be minimized.

I think enough things link to asset for this to be an exception. Its pretty small, but doesn't seem worth it to duplicate in each handler either.

schema.graphql
polymath-eric: I was thinking people would look at the asset to check for nft or not.

I guess its a bit easier to query having it inline, but also seems redundant. Most queries for `AssetTransaction` probably still need to grab the name and ticker from `Asset` anyway.

prashantasdeveloper (approved):

PR #178, opened 2023-10-29
feat: 🎸 Remove call and event handlers

prashantasdeveloper:
### Description

SQ now populates all the data using `blockHandlers` only. Within every block, `ExtrinsicSuccess` events and `timestamp.set` extrinsics are skipped and rest are indexed in the postgres database. This was done to decrease the overall sync time of SQ.

### Breaking Changes

NA

### JIRA Link

DA-922

### Checklist


polymath-eric:
Do you know how much time this is going to save?

It seems like we need to switch to event handlers to be able to use SubQuery dictionaries.

src/mappings/mappingHandlers.ts
```diff
@@ -1,277 +1,99 @@
...
+  for (const ext of wrappedExtrinsics) {
+    if (ext.extrinsic.isSigned) {
+      countExtrinsicsSigned++;
     } else {
-      // for metadata < v14
-      type = genericEvent.meta.args[i].toString();
+      countExtrinsicsUnsigned++;
+    }
+    if (
+      ext.extrinsic.method.section.toLowerCase() !== ModuleIdEnum.timestamp ||
```
polymath-eric, line 87: I was thinking we would use mapping filters in the manifest, but we'd need to switch over to event or call level handlers.

I am not sure how much work this would be. I would think the events have reference to the block.

https://academy.subquery.network/build/manifest/polkadot.html#mapping-handlers-and-filters

polymath-eric (approved):
LGTM, seems like only block handlers is an improvement

prashantasdeveloper:
Upon investigation, I found out that when it comes to block 368, it keeps waiting for new blocks to be created - `INDEXING: Fully synced, waiting for new blocks ` and then starts erring as `2023-11-02T11:04:51.197Z <health> ERROR undefined 
subquery-node_1   | Error: Endpoint is not healthy`.

Is there a way to apply the snapshot on the chain and still keep producing blocks?

PR #179, opened 2023-11-15
feat: 🎸 save customClaimType as entity

sansan:
### Description

- saves CustomClaimType as an Entity
- if Claim has customClaimTypeId -> but that is not yet synced to SQ -> it will be created
- allows querying for all the created CustomClaimType(s)

### JIRA Link

https://polymesh.atlassian.net/browse/DA-909

### Checklist

### DIFF
--- a/src/mappings/entities/mapCustomClaimType.ts
+++ b/src/mappings/entities/mapCustomClaimType.ts
@@ -1,22 +1,7 @@
 import { Codec } from '@polkadot/types/types';
-import { EventIdEnum, ModuleIdEnum, Portfolio, CustomClaimType } from '../../types';
+import { EventIdEnum, ModuleIdEnum, CustomClaimType } from '../../types';
 import { bytesToString, getNumberValue, getTextValue } from '../util';
 import { HandlerArgs } from './common';
-
-export const getPortfolio = async ({
-  identityId,
-  number,
-}: Pick<Portfolio, 'identityId' | 'number'>): Promise<Portfolio> => {
-  const portfolioId = `${identityId}/${number}`;
-
-  const portfolio = await Portfolio.get(portfolioId);
-
-  if (!portfolio) {
-    throw new Error(`Portfolio number ${number} not found for DID ${identityId}`);
-  }
-
-  return portfolio;
-};
 
 export const createCustomClaimType = (
   attributes: Pick<CustomClaimType, 'id' | 'name' | 'identityId'>,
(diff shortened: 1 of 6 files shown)


src/mappings/entities/mapCustomClaimType.ts
```diff
@@ -0,0 +1,105 @@
...
+  const portfolioId = `${identityId}/${number}`;
+
+  const portfolio = await Portfolio.get(portfolioId);
+
+  if (!portfolio) {
+    throw new Error(`Portfolio number ${number} not found for DID ${identityId}`);
+  }
+
+  return portfolio;
+};
```
prashantasdeveloper, line 19: This is not used I guess

src/mappings/entities/mapCustomClaimType.ts
```diff
@@ -0,0 +1,105 @@
...
+  params: Codec[],
+  eventIdx: number
+): Promise<void> => {
+  const [rawDid, rawCustomClaimTypeId, rawName] = params;
+
+  const identityId = getTextValue(rawDid);
+  const id = getNumberValue(rawCustomClaimTypeId);
+  const name = bytesToString(rawName);
+
+  const customClaimType = await Portfolio.get(`${id}`);
```
prashantasdeveloper, line 69:
```suggestion
  const customClaimType = await CustomClaimType.get(`${id}`);
```
Shouldn't this be `CustomClaimType`?

src/mappings/entities/mapCustomClaimType.ts
```diff
@@ -0,0 +1,105 @@
...
+    await createCustomClaimType(
+      {
+        id,
+        name: '',
+        identityId: '',
+      },
+      blockId
+    );
+  }
+};
```
prashantasdeveloper, line 56: I'm not sure why we need this. If the data is not there for old block, we need to use the migration handler to process all the old events and add entries for those and update the current data as well.

Also, creating a CustomClaimType with name and identity as blank would lead to misleading info for the user querying it.

src/mappings/entities/mapCustomClaimType.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+  } else {
+    Object.assign(customClaimType, {
+      name,
+      eventIdx,
+      updatedBlockId: blockId,
+    });
+
+    await customClaimType.save();
+  }
```
prashantasdeveloper, line 88: Is it possible to have `CustomClaimTypeAdded` event being triggered multiple times for the same id? If not, the update case will never be invoked and is not necessary.

prashantasdeveloper (changes requested):
You need to add in code for populating old data using the migration handlers as well. Should be pretty straight forward though

src/mappings/entities/mapCustomClaimType.ts
polymath-eric: SQ is supposed to have perfect serializability just like the chain. When processing a block in SQ its safe to assume every previous block has been processed already.

polymath-eric (commented):
LGTM. Not sure why the tests are failing, looks like Prashant's PR has the same issue though

prashantasdeveloper (approved):

PR #188, opened 2023-12-12
feat: 🎸 Add dictionary support for SQ

prashantasdeveloper:
### Description

Adds in support to use SubQL dictionary

### Breaking Changes

This should require a full re-sync

### JIRA Link

DA-943

### Checklist


polymath-eric (approved):
I think there should be a README update around dictionary, but might be better to add in later after we get it working.

PR #194, opened 2024-01-17
fix: 🐛 remove endBlock from project template

polymath-eric:
### Description

endblock causes sync to stop at a block, which we do not want for our use case

Reference: https://github.com/subquery/documentation/pull/443/files#diff-60836b1220c109ec8d1a41d017d8e7d4d2b56dfa67140099a433e08907ba7626R226

### Breaking Changes

None

### JIRA Link

None

### Checklist

### DIFF
--- a/project.template.ts
+++ b/project.template.ts
@@ -88,7 +88,6 @@ const project: SubstrateProject = {
     {
       kind: SubstrateDatasourceKind.Runtime,
       startBlock: 1,
-      endBlock: 1,
       mapping: {
         file: './dist/index.js',
         handlers: [


project.template.ts
```diff
@@ -88,7 +88,6 @@ const project: SubstrateProject = {
     {
       kind: SubstrateDatasourceKind.Runtime,
       startBlock: 1,
-      endBlock: 1,
```
prashantasdeveloper, line 91: `endBlock` was used to make sure the genesis migration is run only once. 
I used the reference from here(https://academy.subquery.network/faqs/faqs.html#inserting-seed-data-at-project-initiation) to add this in.

polymath-eric:
need for genisis data

PR #195, opened 2024-01-17
feat: 🎸 confidential assets

sansan:
### Checklist

- [ ] Updated the Readme.md (if required) ?

PR #196, opened 2024-01-18
feat: 🎸 confidential assets

quinndiggitypolymath:
### Checklist


quinndiggitypolymath:
@prashantasdeveloper , please fix:
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/26c96ad0031240198b2806af6ffab22d416fb1c9
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/8834f4f56456f0defeec5f538ac3b7d259729c49
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/64349ec039133f2dfa8983ca206ae203a485b16d
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/ce670c1ecc1c69fbfdfc428fd02ca2cdee8c5a51
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/30819eeeaf33c761b6d3135fb7d649eeecf9fd33
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/13f33b6dd199af3e19d708e3498f7b8ed3b820bd
- https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/652a4597fc7dffb652206d38d7184d6dfa03c95f

polymath-eric (approved):
lgtm

polymath-eric:
seems like this version is having trouble starting up in CI

In middleware tests there seems to be an issue with v11+ starting up.

@prashantasdeveloper any ideas: https://github.com/PolymeshAssociation/polymesh-subquery/actions/runs/7564618656/job/20599023280?pr=196

prashantasdeveloper:
@quinndiggitypolymath we need to remove [41b2779](https://github.com/PolymeshAssociation/polymesh-subquery/pull/196/commits/41b2779aff2bb8f78fc535e05cd90fc6fdfbe889) from here. That is a work in progress https://github.com/PolymeshAssociation/polymesh-subquery/pull/195

prashantasdeveloper:
> @prashantasdeveloper any ideas: https://github.com/PolymeshAssociation/polymesh-subquery/actions/runs/7564618656/job/20599023280?pr=196

@polymath-eric IMO the confidential assets one was a WIP and might be causing the failure.

src/mappings/entities/mapConfidentialAccountCreated.ts
```diff
@@ -0,0 +1,26 @@
...
+export const mapConfidentialAccountCreated = async (args: HandlerArgs): Promise<void> => {
+  const { blockId, moduleId, eventId, params } = args;
+
+  if (moduleId !== ModuleIdEnum.confidentialasset) {
+    return;
+  }
+
+  if (eventId !== EventIdEnum.AccountCreated) {
+    return;
+  }
```
prashantasdeveloper, line 14:
```suggestion
  if (moduleId !== ModuleIdEnum.confidentialasset || eventId !== EventIdEnum.AccountCreated) {
    return;
  }
```
nit -> both can be merged

src/mappings/entities/mapConfidentialAsset.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  asset.venueFiltering = enabled;
+  asset.updatedBlockId = blockId;
+
+  await asset.save();
+};
+
+const handleVenuesAllowed = async (blockId: string, params: Codec[]): Promise<void> => {
+  const assetId = getTextValue(params[1]);
+  const venueId = getTextValue(params[2]);
```
prashantasdeveloper, line 62: It seems that `venueId` is param of type Vec<u64>. Can you double check?

src/mappings/entities/mapConfidentialAsset.ts
```diff
@@ -0,0 +1,134 @@
...
+    asset.allowedVenuesId = [venueId];
+  }
+  asset.updatedBlockId = blockId;
+
+  await asset.save();
+};
+
+const handleVenuesBlocked = async (blockId: string, params: Codec[]): Promise<void> => {
+  const assetId = getTextValue(params[1]);
+  const venueId = getTextValue(params[2]);
```
prashantasdeveloper, line 77: It seems that venueId is param of type Vec. Can you double check?

src/mappings/entities/mapConfidentialAsset.ts
```diff
@@ -0,0 +1,134 @@
...
+  const asset = await ConfidentialAsset.get(assetId);
+
+  const index = asset.allowedVenuesId?.indexOf(venueId);
+
+  if (index && index !== -1) {
+    asset.allowedVenuesId?.splice(index, 1);
+    asset.updatedBlockId = blockId;
+
+    await asset.save();
+  }
```
prashantasdeveloper, line 88: If venue is an array list, this needs to be updated

src/mappings/entities/mapConfidentialAssetTransaction.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (moduleId !== ModuleIdEnum.confidentialasset) {
+    return;
+  }
+
+  if (
+    [
+      EventIdEnum.AccountDeposit,
+      EventIdEnum.AccountWithdraw,
+      EventIdEnum.AccountDepositIncoming,
```
prashantasdeveloper, line 79: The way incoming is handled is different. Incoming is just a sign that amount will be deposited. It actually doesn't mean that amount is added. It would be great if you could confirm the working with Robert

prashantasdeveloper (changes requested):

src/mappings/entities/mapConfidentialAsset.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const handleVenuesAllowed = async (blockId: string, params: Codec[]): Promise<void> => {
+  const assetId = getTextValue(params[1]);
+  const venuesAllowed = JSON.parse(params[2].toString()).map(val => getNumberValue(val));
+
+  const asset = await ConfidentialAsset.get(assetId);
+
+  if (asset.allowedVenuesId) {
+    asset.allowedVenuesId = [...asset.allowedVenuesId, ...venuesAllowed];
```
polymath-eric, line 67: Does this need deduplication? I think the chain allows set to be triggered even though no venues change.

Doesn't look as nice, but might be safer:
```ts
...new Set([...allowedVenues, ...venuesAllowed])
```

schema.graphql
```diff
@@ -2329,3 +2358,120 @@ type CustomClaimType @entity {
...
+  Affirmed
+  Rejected
+}
+
+enum TransactionStatusEnum {
+  Created
+  Executed
+  Rejected
+}
+
```
polymath-eric, line 2378: """
Represents a Confidential Transaction sender proof of funds
"""

And comments for the other types. Not sure how much they add, but it will look more consistent

polymath-eric (commented):
overall lgtm.

Could be nice to squash some of the refactor commits. I noticed there was already a force push to this branch.

schema.graphql
```diff
@@ -2329,3 +2358,157 @@ type CustomClaimType @entity {
...
+"""
+Represents a confidential asset holder
+"""
+type ConfidentialAssetHolder @entity {
+  id: ID!
+  account: ConfidentialAccount!
+  asset: ConfidentialAsset!
+  amount: String
```
prashantasdeveloper, line 2445:
```suggestion
  """
  the encrypted amount 
  """
  amount: String 
```

schema.graphql
```diff
@@ -2329,3 +2358,157 @@ type CustomClaimType @entity {
...
+"""
+Represents a confidential asset transaction affirmation
+"""
+type ConfidentialTransactionAffirmation @entity {
+  id: ID!
+  transaction: ConfidentialTransaction
+  leg: Int
+  account: ConfidentialAccount
+  type: AffirmingPartyEnum
+  status: AffirmStatusEnum
```
prashantasdeveloper, line 2459: Won't all these properties be mandatory (i.e. needing `!`)

schema.graphql
```diff
@@ -2329,3 +2358,157 @@ type CustomClaimType @entity {
...
+Represents a leg of a confidential asset transaction
+"""
+type ConfidentialLeg @entity {
+  id: ID!
+  transaction: ConfidentialTransaction!
+  sender: ConfidentialAccount!
+  receiver: ConfidentialAccount!
+  mediators: [String!]
+  auditors: [String!]
+  asset: ConfidentialAsset
```
prashantasdeveloper, line 2475:
```suggestion
  asset: ConfidentialAsset!
```
Was this intentional? Can we have a leg without asset?

schema.graphql
```diff
@@ -2329,3 +2358,157 @@ type CustomClaimType @entity {
...
+  memo: String
+  pendingAffirmations: Int!
+  createdBlock: Block!
+  updatedBlock: Block!
+}
+
+"""
+Represents a confidential asset transaction
+"""
+type ConfidentialAssetTransaction @entity {
```
prashantasdeveloper, line 2500:
```suggestion
type ConfidentialAssetHistory @entity {
```
The name would be a bit confusing to users. Maybe we can add `history` as suffix?

schema.graphql
```diff
@@ -2329,3 +2358,157 @@ type CustomClaimType @entity {
...
+Used for many to many relation between ConfidentialAsset and ConfidentialVenue
+"""
+type ConfidentialAssetAllowedVenue @entity {
+  id: ID!
+  asset: ConfidentialAsset!
+  venue: ConfidentialVenue!
+  isAllowed: Boolean!
+  createdBlock: Block!
+  updatedBlock: Block!
+}
```
prashantasdeveloper, line 2419: I don't think we need a separate entity for this. We can have `allowedVenues: [Int!]` within ConfidentialAsset itself.

src/mappings/entities/mapConfidentialAsset.ts
```diff
@@ -0,0 +1,217 @@
...
+  await Promise.all([...foundAllowedVenuesUpdatePromises, ...blockedVenuesPromises]);
+
+  const asset = await ConfidentialAsset.get(assetId);
+
+  if (asset) {
+    asset.updatedBlockId = blockId;
+
+    await asset.save();
+  }
+};
```
prashantasdeveloper, line 170: The logic can be simplified here.

Note - For venues allowed/blocked, the venue doesn't have to exists (For example you can allow say venue id 9999 even if doesn't exists. So we can only track the IDs here. That is why I recommended to remove the entity and only add a single attribute as `allowedVenues` to `ConfidentialAsset`. 

Also, if we block a venue say 45 and it is not already allowed, nothing happens (list remains unaltered). In short, we only keep a track of allowed venues and if the blocked venues are a part of allowed we remove that from the list, otherwise do nothing

src/mappings/entities/mapConfidentialAssetTransaction.ts
```diff
@@ -0,0 +1,105 @@
...
+    existingAssetHolder.amount = balance;
+    existingAssetHolder.updatedBlockId = blockId;
+
+    await existingAssetHolder.save();
+  } else {
+    await ConfidentialAssetHolder.create({
+      id,
+      accountId,
+      assetId,
+      amount,
```
prashantasdeveloper, line 52:
```suggestion
      amount: balance,
```
Just for uniformity (that we update balance here), even though for the first time both will be same

src/mappings/entities/mapConfidentialTransaction.ts
```diff
@@ -0,0 +1,220 @@
...
+  const transaction = await ConfidentialTransaction.get(transactionId);
+
+  if (transaction) {
+    transaction.status = TransactionStatusEnum.Rejected;
+    transaction.updatedBlockId = blockId;
+    transaction.memo = memo;
+
+    await transaction.save();
+  }
+};
```
prashantasdeveloper, line 122: Both of these can be merged, extracting a status variable as method parameter

prashantasdeveloper (changes requested):

schema.graphql
```diff
@@ -2329,3 +2358,154 @@ type CustomClaimType @entity {
...
+"""
+type ConfidentialTransactionAffirmation @entity {
+  id: ID!
+  transaction: ConfidentialTransaction!
+  leg: Int!
+  account: ConfidentialAccount
+  identity: Identity!
+  type: AffirmingPartyEnum!
+  status: AffirmStatusEnum!
+  proofs: [SenderProof]
```
polymath-eric, line 2453:
```suggestion
  proofs: [SenderProof!]
```
The array is optional, but if its present, then the element isn't. Although it could still be empty

I am not sure, but that would be consistent with `[AssetAuditorGroup!]` below.

schema.graphql
```diff
@@ -2329,3 +2358,154 @@ type CustomClaimType @entity {
...
+  updatedBlock: Block!
+}
+
+"""
+Represents a confidential asset transaction affirmation
+"""
+type ConfidentialTransactionAffirmation @entity {
+  id: ID!
+  transaction: ConfidentialTransaction!
+  leg: Int!
```
polymath-eric, line 2448:
```suggestion
  legId: Int!
```
I think just plain "leg" should refer to an entity

schema.graphql
```diff
@@ -2329,3 +2358,154 @@ type CustomClaimType @entity {
...
+
+"""
+Represents a leg of a confidential asset transaction
+"""
+type ConfidentialLeg @entity {
+  id: ID!
+  transaction: ConfidentialTransaction!
+  sender: ConfidentialAccount!
+  receiver: ConfidentialAccount!
+  mediators: [String!]
```
polymath-eric, line 2471:
```suggestion
  mediators: [String!]!
```
mediators might be empty, and in that case an empty array saved, but the field should never be nullable, and each element is non nullable as well

schema.graphql
```diff
@@ -2329,3 +2358,154 @@ type CustomClaimType @entity {
...
+  assetAuditors: [AssetAuditorGroup!]
```
polymath-eric, line 2472:
```suggestion
  assetAuditors: [AssetAuditorGroup!]!
```
empty array should be saved instead of `null` or `undefined` imo

schema.graphql
```diff
@@ -2329,3 +2358,154 @@ type CustomClaimType @entity {
...
+Represents a confidential transaction
+"""
+type ConfidentialTransaction @entity {
+  id: ID!
+  venue: ConfidentialVenue! @index(unique: false)
+  status: ConfidentialTransactionStatusEnum!
+  eventId: EventIdEnum!
+  eventIdx: Int!
+  legs: [ConfidentialLeg!]! @derivedFrom(field: "transaction")
+  affirmations: [ConfidentialTransactionAffirmation] @derivedFrom(field: "transaction")
```
polymath-eric, line 2487: could this use some non nullable `!` marks?

polymath-eric (commented):
some comments around possible `!` in the schema.

The way I see it is every `!` declared here is one less code branch in every consumer.

PR #200, opened 2024-02-20
feat: 🎸 add tracking for asset required mediators

polymath-eric:
### Description

Tracks required mediators for assets

### Breaking Changes

None

### JIRA Link

✅ Closes: DA-1025

### Checklist

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -2351,3 +2351,18 @@ type CustomClaimType @entity {
   createdBlock: Block!
   updatedBlock: Block!
 }
+
+"""
+Represents a mediator's approval for an instruction
+"""
+type MeditatorAffirmation @entity {
+  id: ID! # did/instruction ID
+  identity: Identity!
+  instruction: Instruction!
+  """
+  If expiry is present the time should be check top ensure the affirmation is still valid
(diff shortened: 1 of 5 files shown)


polymath-eric:
Looks like we have stuff in `alpha-signed` we need into master: https://github.com/PolymeshAssociation/polymesh-subquery/compare/master...alpha-signed

After this PR gets merged I will get alpha-signed branch into master after making sure its OK, and then get rid of it.

schema.graphql
```diff
@@ -309,6 +309,7 @@ enum EventIdEnum {
   RegisterAssetMetadataGlobalType
   MetadataValueDeleted
   LocalMetadataKeyDeleted
+  AssetMediatorsAdded
   SetAssetMediators
```
polymath-eric, line 313: `SetAssetMediators` was on the release notes, but its not actually an event. Its `AssetMediatorsAdded` instead.

Postgres doesn't allow for removal of enums. The entire type needs to be dropped and then recreated.

If resync is fast now maybe its not a big deal to force one to clean up

VictorVicente (approved):

PR #211, opened 2024-03-19
feat: 🎸 Enable historical state for SQ

prashantasdeveloper:
### Description

Enable historical state for SQ. This will allow querying the state of SQ at any block height

### Breaking Changes

This will require a full resync, since enabling this will re-index all the entities with their historical state. If you wish to disable the historical state tracking and continue with old format of indexing, pass `--disable-historical=true` as
command argument for the subquery node container.

### JIRA Link

DA-1126

### Checklist


polymath-eric (commented):
Any idea why it broke the snapshot tests?

The nodeId not lining up makes sense to me, but some of them seem like completely different addresses/identities now.

VictorVicente (approved):

polymath-eric (commented):
I think more cases need an order by clause, like: https://github.com/PolymeshAssociation/polymesh-subquery/blob/98e938822072ef566e423a57fd68cfa6c2326496/tests/entities/identities.test.ts#L63

PR #230, opened 2024-05-24
feat: 🎸 add params to MultiSigProposal

sansan:
### Description

adds extrinsic params to MultiSigProposal 

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1164

### Checklist

### DIFF
--- a/db/migrations/3_extrinsic_hex_prefix.sql
+++ b/db/migrations/3_extrinsic_hex_prefix.sql
@@ -1,0 +1,3 @@
+UPDATE extrinsics
+SET extrinsic_hash = CONCAT('0x', extrinsic_hash)
+WHERE extrinsic_hash NOT LIKE '0x%';


polymath-eric (approved):

polymath-eric:
@sansan I switched this to draft so its clearer to not merge atm. Feel free to switch it back once testnet is synced

schema.graphql
```diff
@@ -2373,6 +2382,7 @@ type MultiSigProposal @entity {
   datetime: Date!
   eventIdx: Int!
   extrinsicIdx: Int!
+  params: MultiSigProposalParams
```
prashantasdeveloper, line 2385:
```suggestion
  params: MultiSigProposalParams!
```
Shouldn't this be a non-null field? IIRC all proposal are bound to have params

src/mappings/entities/mapMultiSig.ts
```diff
@@ -158,6 +159,21 @@ const handleMultiSigProposalAdded = async (
   const multisigId = getTextValue(rawMultiSigAddress);
   const proposalId = getNumberValue(rawProposalId);
   const creatorAccount = extrinsic?.extrinsic.signer.toString();
+  const parsed = (extrinsic?.extrinsic?.toHuman() as any).method.args || [];
+
+  const module = parsed[1]?.value.call_module;
+  const call = parsed[1]?.value.call_function;
```
prashantasdeveloper, line 165: @polymath-eric do you think we should parse these into our enum types (CallIdEnum and ModuleIdEnum)? That way it would be easily filterable.

src/mappings/entities/mapMultiSig.ts
```diff
@@ -158,6 +159,21 @@ const handleMultiSigProposalAdded = async (
...
+  const expiry = parsed[2]?.value;
+  const args = parsed[1]?.value.call_args;
+
+  const proposalParams: MultiSigProposalParams = {
+    module,
+    call,
+    callIndex,
+    expiry,
+    args,
+  };
```
prashantasdeveloper, line 176: In case of bridge transaction, these won't be there . Check - https://polymesh-testnet.subscan.io/block/12136029?tab=event

prashantasdeveloper (changes requested):

src/mappings/entities/mapMultiSig.ts
polymath-eric: For filterability wouldn't they need to go onto the `MultiSigProposal` itself. 

I thought JSON fields get worse indexing compared to regular columns.

In general I agree it would be better to use the enum value though.

schema.graphql
```diff
@@ -2358,6 +2358,15 @@ enum MultiSigProposalStatusEnum {
   Deleted
 }
 
+
+type MultiSigProposalParams @jsonField {
+  module: String
+  call: String
```
polymath-eric, line 2364: Can use the enum values?

polymath-eric (dismissed):

src/mappings/entities/mapExtrinsic.ts
```diff
@@ -23,7 +22,7 @@ export function createExtrinsic(extrinsic: SubstrateExtrinsic): Extrinsic {
     signedbyAddress: signedbyAddress ? 1 : 0,
     address,
     nonce: extrinsic.extrinsic.nonce.toNumber(),
-    extrinsicHash: hexStripPrefix(extrinsic.extrinsic.hash.toJSON()),
+    extrinsicHash: extrinsic.extrinsic.hash.toJSON(),
```
polymath-eric, line 25: IMO either we leave this without the prefix, or mark it as a breaking change.

It seems weird to me if someone didn't resync to have a mix of prefix and non prefix values in their DB.

I do think its better with the prefix though, but if we are adding it in it should be clear the response changed for people.

prashantasdeveloper: I think this was left by mistake. Even the SDK directly sends 0x prefixed hash value in `getTransactionByHash`, so feels like a fix to me. We can add in migration to update the existing values

polymath-eric: I think this was because harvester didn't prefix them and we had to run tooling with the exact data

I think an update is better, but a migration so its consistent would be good, and it should  be marked as a feature, not as a chore.

polymath-eric (approved):

prashantasdeveloper (dismissed):

db/migrations/3_extrinsic_hex_prefix.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE extrinsics
+SET extrinsic_hash = CONCAT('0x', extrinsic_hash)
+WHERE extrinsic_hash NOT LIKE '0x%';
```
polymath-eric, line 3: Does this PR require a resync so the old params get populated?

If so, then we don't need to update existing data.

PR #231, opened 2024-05-30
refactor: 💡 remove Event/Extrinsic/Block dependency

polymath-eric:
### Description

removes dependency on Event, Extrinsic and Block within the handlers

These were for migrations, which we shouldn't need to run. By making sure there is no accidental coupling to the Blocks/Events we should have more flexibility in how we process.

There is no foreign key constraint on `Blocks` and the query service handles the null case, so if we know there is know dependency then its safe to process them in parallel.

### Checklist


prashantasdeveloper (approved):

polymath-eric:
Skipping the block took sync time for staging 1010260 from 48 seconds to ~33 seconds locally

PR #232, opened 2024-05-31
refactor: 💡 improve sync performance on large batches

polymath-eric:
### Description

skip recreating already indexed extrinsics

locally, reduces staging block 1010260 sync time from ~33 seconds to ~10 seconds.

### Checklist

### DIFF
--- a/src/mappings/entities/mapBlock.ts
+++ b/src/mappings/entities/mapBlock.ts
@@ -1,7 +1,7 @@
 import { SubstrateBlock } from '@subql/types';
 import { Block } from '../../types';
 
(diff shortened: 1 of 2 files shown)


src/mappings/entities/mapBlock.ts
```diff
@@ -1,7 +1,7 @@
 import { SubstrateBlock } from '@subql/types';
 import { Block } from '../../types';
 
-export const mapBlock = async (block: SubstrateBlock): Promise<Block> => {
```
polymath-eric, line 4: this doesn't make a detectable difference, but I figure non promise is faster than promise

VictorVicente (approved):

PR #234, opened 2024-06-03
refactor: 💡 make getEventParams sync + add genesisHandler

polymath-eric:
### Description

avoid unnecessary promises and move some handlers to be ran on a per block basis. Also adds an explicit `handleGenesis` function

### Checklist


polymath-eric:
Nothing here that impacts performance in a measurable way, but might make sync a little bit faster.

VictorVicente (approved):

PR #235, opened 2024-06-03
refactor: 💡 remove data migration handler

polymath-eric:
### Description

removes data migration handler to simplify logical flow. The handler was did not handle the case for historical state.

Also migrates upgrade handler to a dedicated function

### Checklist


src/mappings/mappingHandlers.ts
```diff
@@ -6,39 +6,41 @@ import { createExtrinsic } from './entities/mapExtrinsic';
...
+   */
+  await mapSubqueryVersion().catch(e => logError(e));
+}
+
 export async function handleEvent(substrateEvent: SubstrateEvent): Promise<void> {
-  const header = substrateEvent.block.block.header;
-  const blockId = header.number.toNumber();
+  if (!startupHandled) {
+    await handleStartup();
+    startupHandled = true;
```
prashantasdeveloper, line 35: I noticed that, from the time we stopped adding semantic bot commits the version in `package.json` doesn't get updated. So we need to either update that manually or keep back the git commits from bot to make sure the version is updated correctly. 

We use this version as MIN_SQ_VERSION in sdk to check if the SQ is compatible with SDK functions.

prashantasdeveloper (approved):
Left a comment around package version, that can be done in separate task. This one LGTM

src/mappings/mappingHandlers.ts
polymath-eric: makes sense. I'll bump it when I am ready to promote all of the alpha optimizations to master

PR #259, opened 2024-10-16
Add new events for 7.x chain and some other fixes

prashantasdeveloper:
### Description

- Portfolio movements incorrectly had ticker value populated against asset ID field for Non Fungible tokens. This fixes the issue. (DA-1327)
- Adds new events and extrinsics from latest 7.x chain (DA-1333)
- Add new handler for mapping event `TickerUnlinkedFromAsset` (DA-1333)
- Update logic of asset ID generation for legacy tickers (to generate valid UUID) (DA-1333)
- Bump `@polymeshassociation/polymesh-types` to `5.14.0` (DA-1331)

### Breaking Changes

NA

### JIRA Link

DA-1327, DA-1333, DA-1331

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "polymesh-subquery",
-  "version": "16.0.1",
+  "version": "16.1.0",
   "author": "Polymesh Association",
   "license": "Apache-2.0",
   "description": "A Polymesh Chain Indexer, providing a GraphQL interface",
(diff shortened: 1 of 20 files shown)


prashantasdeveloper:
I haven't added the migration because next change with updated assetID calculation using genesis hash requires full re-sync. Just FYI query to fix the data would have been - 

```
update portfolio_movements pm 
set asset_id = a.id
from assets a 
where a.ticker = pm.asset_id;
```

VictorVicente (dismissed):
Shouldn't it be related to some unit test?

VictorVicente (approved):

PR #261, opened 2024-10-24
feat: 🎸 Make `Distribution.currency` of type `Asset`

prashantasdeveloper:
### Description

Make `Distribution.currency` of type `Asset`

### Breaking Changes

🧨 Type of `currency` in `Distribution` entity has been changed to `Asset` type

### JIRA Link

DA-1353

### Checklist

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -2037,7 +2037,7 @@ type Distribution @entity {
   localId: Int! @index(unique: false)
   asset: Asset! @index(unique: false)
   portfolio: Portfolio!
-  currency: String!
+  currency: Asset!
   perShare: BigInt!
   amount: BigInt!
   remaining: BigInt!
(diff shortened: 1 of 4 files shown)


schema.graphql
```diff
@@ -2037,7 +2037,7 @@ type Distribution @entity {
   localId: Int! @index(unique: false)
   asset: Asset! @index(unique: false)
   portfolio: Portfolio!
-  currency: String!
```
polymath-eric, line 2040: Is this always an Asset? Or can people put in a bad ID? I forget if the chain validates for the existence of it or not

polymath-eric (approved):
LGTM, but are you sure currency is always an existing asset?

schema.graphql
polymath-eric: I guess distribution should be safe. I was thinking of offering

PR #265, opened 2024-11-05
fix: 🐛 claim scopes with tickers not being revoked

polymath-eric:
### Description

use asset ID in scoped claims to prevent issue where they would not be revoked when a ticker was linked

### Checklist


db/migrations/3_claim_ticker_to_id.sql
```diff
@@ -0,0 +1,52 @@
+-- NOTE: This migration should only run after all pre 7.0 blocks have been processed
+-- Similar logic should be triggered when the 7.0 upgrade is reached
+
```
prashantasdeveloper, line 3: Migration are supposed to be idempotent whenever they get executed. We never know when one decides they should run. What if someone starts the sync from 16.0.0 (which is compatible with dual version) and then bumps it in between.

db/migrations/3_claim_ticker_to_id.sql
```diff
@@ -0,0 +1,52 @@
...
+WITH spec AS (
+    SELECT spec_version_id as version
+    FROM blocks
+    ORDER BY datetime DESC
+    LIMIT 1
+)
+UPDATE claims
+SET id =
+    REGEXP_REPLACE(id, '/[^/]+$', '/' || (scope->>'assetId')::TEXT),  -- Replace text after the last '/'
+    scope = jsonb_set(scope, '{value}', to_jsonb(scope->>'assetId'))
```
prashantasdeveloper, line 14: `value` and `assetId` were both kept to support dual version. If you update the value to `assetId` we would require changes in SDK side to return ticker value for old claims.

db/migrations/3_claim_ticker_to_id.sql
```diff
@@ -0,0 +1,52 @@
...
+    FROM blocks
+    ORDER BY datetime DESC
+    LIMIT 1
+),
+revoked_events AS (
+    SELECT
+           event_arg_1::json AS event_arg_json,
+           event_arg_0 || '/' ||
+           (SELECT json_object_keys(event_arg_1::json->'claim') LIMIT 1) || '/Asset/' ||
+           (event_arg_1::json->'claim'->(SELECT json_object_keys(event_arg_1::json->'claim') LIMIT 1)->>'Asset')
```
prashantasdeveloper, line 30: It is not mandatory that claim revoked is always of type that has asset scope. What if CustomerDueDiligence or NoData type claim is revoked? I saw in staging data some of these instances to be there.

db/migrations/3_claim_ticker_to_id.sql
```diff
@@ -0,0 +1,52 @@
...
+           AS claim_id,
+           event_arg_1::json->>'issuanceDate' as revoke_date FROM events
+    WHERE event_id = 'ClaimRevoked'
```
prashantasdeveloper, line 33:
```suggestion
    select
    concat(
      e.event_arg_0,
      '/',
      e.claim_type,
      case 
        when e.claim_type = 'CustomerDueDiligence' 
          then concat('/', attributes->1->'value'->'claim'->>'CustomerDueDiligence')
        when e.claim_type = 'NoData' 
          then ''
        else 
          concat(
            case 
              when e.claim_type = 'Custom' 
              then e.attributes->1->'value'->'claim'->'Custom'->>'col1' 
              else '' 
            end, 
            case 
              when e.claim_scope is not null 
              then  
                case 
                  when (e.claim_scope::jsonb)->>'type' is not null 
                  then 
                    concat(
                      '/', 
                      case 
                        when (e.claim_scope::jsonb)->>'type' = 'Ticker' 
                        then 'Asset'
                        else (e.claim_scope::jsonb)->>'type'
                      end,
                      '/', 
                      case 
                        when (e.claim_scope::jsonb)->>'type' = 'Ticker' 
                        then a.id
                        else (e.claim_scope::jsonb)->>'value'
                      end
                    )
                  else '//' 
                end
              else ''
            end,
            case 
              when e.claim_type = 'Jurisdiction' 
              then concat('/', e.attributes->1->'value'->'claim'->'Jurisdiction'->>'col1') 
              else '' 
            end
          )
      end
    ) as claim_id,
    e.attributes->1->'value'->'issuanceDate' as revoke_date,
    e.block_id,
    e.event_idx
  from events e 
  left join assets a 
    on 
      case 
        when (e.claim_scope::jsonb)->>'type' = 'Ticker' 
        then a.ticker = (e.claim_scope::jsonb)->>'value' 
        else false 
      end
  where e.event_id = 'ClaimRevoked'
```

db/migrations/3_claim_ticker_to_id.sql
```diff
@@ -0,0 +1,52 @@
...
+),
+claims_to_update AS (
+    SELECT
+        lc.id AS claim_id,
+        re.revoke_date
+    FROM revoked_events re
+    JOIN latest_claims lc ON re.claim_id = lc.id
+)
+UPDATE claims
+SET revoke_date = claims_to_update.revoke_date::numeric
```
prashantasdeveloper, line 48: update the `updated_block_id` as well

src/mappings/entities/block/mapChainUpgrade.ts
```diff
@@ -29,6 +30,11 @@ export default async (substrateEvent: SubstrateEvent): Promise<void> => {
       logger.info(`Transaction version was not changed for the chain upgrade`);
     }
 
+    if (specVersion >= 7000000 && specVersion <= 7000999) {
+      logger.info('v7 upgrade detected - migrating claim tickers to asset IDs');
+      await v7ClaimIdHandler(substrateEvent);
+    }
+
```
prashantasdeveloper, line 37: we should avoid doing as many actions at chain upgrade. If we just fix the id of claims to always use assetId, this won't be needed

src/mappings/entities/identities/mapClaim.ts
```diff
@@ -71,6 +71,23 @@ const getId = (
...
+const processClaim = async (claimScope: any, block: SubstrateBlock): Promise<Scope> => {
+  const scope = JSON.parse(claimScope);
+
+  if (scope.type === ClaimScopeTypeEnum.Ticker || scope.type === ClaimScopeTypeEnum.Asset) {
+    scope.type = ClaimScopeTypeEnum.Asset;
+    const { assetId, ticker } = await getAssetIdWithTicker(scope.value, block);
+
+    if (block.specVersion < 7000000) {
+      scope.value = ticker;
+    }
```
prashantasdeveloper, line 83: always setting this value is no harm. Before specVersion 7000000, ticker value will always be there and for cases after 7000000 where ticker is not linked to asset, value will remain null. If you want to avoid setting `value: undefined` maybe the check can be modified as 
```suggestion
    if (ticker) {
      scope.value = ticker;
    }
```
and add comment that value can be null  if ticker is not liked with asset

src/mappings/migrations/v7ClaimIdHandler.ts
```diff
@@ -0,0 +1,47 @@
+import { SubstrateEvent } from '@subql/types';
+import { getManager } from 'typeorm';
+
+const rawSql = `
+UPDATE claims
+SET id =
```
prashantasdeveloper, line 6: this can be removed once above comments are resolved

prashantasdeveloper (changes requested):

src/mappings/entities/block/mapChainUpgrade.ts
polymath-eric: The SQL doesn't work either since the process gets sandboxed in SQ.

I should be able to update just IDs using the ORM though.

src/mappings/entities/identities/mapClaim.ts
polymath-eric: With ticker linked/unlinked events I didn't want these to come out of sync.

I'll continue setting it here for now since I think the entity refactor story normalizes the DB to make it not a concern.

db/migrations/3_claim_ticker_to_id.sql
polymath-eric: If the start with v16 there are two cases, where they update before syncing v7 upgrade and after.

If before, then the v7 block upgrade code will do the needed updates (only IDs are needed like you pointed out).

If after, then this migration will take care of it.

The `WITH spec AS` subquery is supposed to grab the latest block processed and be checked in the updates to handle this.

db/migrations/3_claim_ticker_to_id.sql
polymath-eric: The non scoped claims are correctly revoked with the current handler, so they don't need to be updated. Good catch with the other types though.

src/mappings/entities/block/mapChainUpgrade.ts
```diff
@@ -29,6 +31,11 @@ export default async (substrateEvent: SubstrateEvent): Promise<void> => {
       logger.info(`Transaction version was not changed for the chain upgrade`);
     }
 
+    if (specVersion >= 7000000 && specVersion <= 7000999) {
+      logger.info('v7 upgrade detected - migrating claim tickers to asset IDs');
+      await v7UpdateClaimIds();
+    }
+
```
prashantasdeveloper, line 38: I still think we should get rid of this

src/mappings/entities/identities/mapClaim.ts
```diff
@@ -57,7 +57,7 @@ const getId = (
   if (scope) {
     // Not applicable in case of CustomerDueDiligence, InvestorUniquenessV2Claim, NoData claim types
     idAttributes.push(scope.type);
-    idAttributes.push(scope.value);
+    idAttributes.push(scope.assetId ?? scope.value);
```
prashantasdeveloper, line 60: Since we have added this, resync will make sure that assetID is always used. If we update the existing asset scope types to use assetId  irrespective of spec version, we won't need to handle chain upgrade

src/mappings/entities/identities/mapClaim.ts
```diff
@@ -179,3 +189,24 @@ const handleScopes = async (
...
+    if (claim?.scope?.assetId) {
+      updateClaims.push(Claim.remove(claim.id)); // the old record needs to be "removed" since the new record as a new ID
+
+      claim.id = claim.id.replace(/(Asset\/)[^/]+/, `$1${claim.scope.assetId}`);
+      updateClaims.push(claim.save());
+    }
+  });
+  await Promise.all(updateClaims);
+  logger.info('updated claim IDs with AssetId');
+};
```
prashantasdeveloper, line 212: If we remove the spec version check in migration, this is not needed

prashantasdeveloper (changes requested):

src/mappings/entities/block/mapChainUpgrade.ts
prashantasdeveloper: That won't be necessary, we can handle fixing the ID of the claims entity to always use assetId from scope and voila, we don't need to do anything

db/migrations/3_claim_ticker_to_id.sql
polymath-eric: Makes sense. I was thinking there was no such thing as asset IDs on 6.0 for some reason, but guess thats not the case with the calculate legacy for ticker function.

docker-compose.yml
```diff
@@ -35,17 +35,17 @@ services:
...
       # NETWORK_ENDPOINT: wss://staging-rpc.polymesh.dev/
       # NETWORK_CHAIN_ID: '0x3c3183f6d701500766ff7d147b79c4f10014a095eaaa98e960dcef6b3ead50ee'
       # NETWORK_DICTIONARY: https://subql-dictionary.polymesh.dev
       # Testnet
-      # NETWORK_ENDPOINT: wss://testnet-rpc.polymesh.live
-      # NETWORK_CHAIN_ID: '0x2ace05e703aa50b48c0ccccfc8b424f7aab9a1e2c424ed12e45d20b1e8ffd0d6'
-      # NETWORK_DICTIONARY: https://dev.polymesh.tech/dictionary/testnet
+      NETWORK_ENDPOINT: wss://testnet-rpc.polymesh.live
+      NETWORK_CHAIN_ID: '0x2ace05e703aa50b48c0ccccfc8b424f7aab9a1e2c424ed12e45d20b1e8ffd0d6'
+      NETWORK_DICTIONARY: https://testnet-subql-dictionary.polymesh.live/
```
prashantasdeveloper, line 48: this should be reverted

prashantasdeveloper (dismissed):

prashantasdeveloper (approved):

PR #272, opened 2024-11-15
feat: 🎸 pad block id to provide expected ordering

polymath-eric:
Draft note: Either a migration is needed, or needing a resync will need to be marked

### Description

allow ordering by block ID by padding it with 0 such that lexical order becomes numeric order

### JIRA Link

✅ Closes: DA-1371

### Checklist


prashantasdeveloper (changes requested):
We can reduce so many changes by just updating one single method + couple of more places I guess where Block entity is updated. Also, instead of writing a migration, we should mark this as breaking change and get rid of existing migrations as well. 

Also, there is the created_at column which we can get rid of now. We just need to make sure that SDK is also updated to not use the created_at column. Or maybe we can drop this in next change moving forward

prashantasdeveloper:
While we are at it, please do pad the instruction ID, authorization ID as well

polymath-eric:
@prashantasdeveloper 
> While we are at it, please do pad the instruction ID, authorization ID as well

I made the changes but I am thinking we should just leave those IDs as they are. The SDK would need to be updated which would slow down the release. Also if people want instructions or auths in order the could just use the block ID like for any other entity.

polymath-eric:
> Also, there is the created_at column which we can get rid of now. We just need to make sure that SDK is also updated to not use the created_at column. Or maybe we can drop this in next change moving forward

Dropping it becomes a breaking change. I agree we should clean it up eventually though.

db/migrations/4_pad_id.sql
```diff
@@ -0,0 +1,82 @@
...
+            AND column_name = 'updated_block_id'
+        ) THEN
+            EXECUTE format(
+                'UPDATE %I SET updated_block_id = LPAD(updated_block_id, 10, ''0'') WHERE updated_block_id IS NOT NULL;',
+                tbl_name
+            );
+        END IF;
+    END LOOP;
+END $$;
+
```
prashantasdeveloper, line 45: Padding of `id` field in Block entity is missing

schema.graphql
```diff
@@ -1696,7 +1696,8 @@ enum AuthorizationStatusEnum {
 Represents a request to grant an Account some permission to perform actions for an Identity. e.g. become a Portfolio custodian
 """
 type Authorization @entity {
-  id: ID!
+  id: ID! # Left padded AuthID
+  authId: Int! # AuthId as Int
```
prashantasdeveloper, line 1700:
```suggestion
  authId: Int! @index(unique: true) # AuthId as Int
```

schema.graphql
```diff
@@ -1878,7 +1879,8 @@ If any party fails to satisfy any compliance or transfer restriction the Instruc
 and no movement of assets will occur.
 """
 type Instruction @entity {
-  id: ID! #instructionId
+  id: ID! # instructionId, left padded with 0
+  instructionId: Int! # instructionId as Int
```
prashantasdeveloper, line 1883:
```suggestion
  instructionId: Int! @index(unique: true) # instructionId as Int
```

db/migrations/4_pad_id.sql
```diff
@@ -0,0 +1,82 @@
...
+    FOR tbl_name IN
+        SELECT table_name
+        FROM information_schema.tables
+        WHERE table_schema = current_schema_name AND table_type = 'BASE TABLE'
+    LOOP
+        -- Check if `created_block_id` column exists and update values
+        IF EXISTS (
+            SELECT 1
+            FROM information_schema.columns
+            WHERE table_schema = current_schema_name
```
prashantasdeveloper, line 20: IMO we should re-sync all of these. Just updating only these value won't do the trick, we will have to make sure all the composite ID's of other entities (where ID for example is created as `1000/1` [`blockId/eventIdx`]) will need to be updated as well, to keep the data correct with/without sync.

Feels like a lot of pain to go and add queries for all such entities.

prashantasdeveloper (changes requested):
I left comment about removing migrations and forcing re-sync here. That will also allow us to add dual version SQ support in SDK for this. 

Also, you forgot to update a few places where block ID is used. `src/utils/assets.ts` (Search for 'Block.get'), `mapBlock`, `mapChainUpgrade`, `mapExtrinsics`, `mapEvent` (search for 'header.number'). Also in `genesisHandler` while inserting block 0.

prashantasdeveloper:
> Dropping it becomes a breaking change. I agree we should clean it up eventually though.

If we are deciding to go on resync for this change, we should drop created_at as well

prashantasdeveloper:
> @prashantasdeveloper
> 
> > While we are at it, please do pad the instruction ID, authorization ID as well
> 
> I made the changes but I am thinking we should just leave those IDs as they are. The SDK would need to be updated which would slow down the release. Also if people want instructions or auths in order the could just use the block ID like for any other entity.

Anyhow, we will be needing to add in dual version SQ support in SDK. So it makes sense to add in as many breaking changes as possible

src/mappings/entities/block/mapChainUpgrade.ts
```diff
@@ -24,7 +25,8 @@ export default async (substrateEvent: SubstrateEvent): Promise<void> => {
 
     if (txVersion !== transactionVersion) {
       logger.info(`Major chain upgrade found with transaction version upgraded `);
-      await handleMultiSigProposalDeleted(substrateEvent.block.block.header.number.toString());
+      const blockId = padId(substrateEvent.block.block.header.number.toString());
+      await handleMultiSigProposalDeleted(blockId);
```
prashantasdeveloper, line 29: On rebasing, this will change.

src/mappings/entities/block/mapExtrinsic.ts
```diff
@@ -1,17 +1,17 @@
...
+  const blockId = padId(extrinsic.block.block.header.number.toString());
   const extrinsicIdx = extrinsic.idx;
   const signedbyAddress = !extrinsic.extrinsic.signer.isEmpty;
   const address = signedbyAddress ? extrinsic.extrinsic.signer.toString() : null;
   const paramsTxt = JSON.stringify((extrinsic.extrinsic.toHuman() as any).method.args);
 
   return Extrinsic.create({
     id: `${blockId}/${extrinsicIdx}`,
-    blockId,
+    blockId: blockId,
```
prashantasdeveloper, line 14:
```suggestion
    blockId,
```

prashantasdeveloper (dismissed):
LGTM

prashantasdeveloper (approved):

PR #290, opened 2025-01-08
fix: 🐛 Correct handler mapping for `TransactionFeePaid` event

prashantasdeveloper:
BREAKING CHANGE: 🧨 Since the entries for `TransactionFeePaid` event where not indexed, this will require a full re-sync

### Description

Correct handler mapping for `TransactionFeePaid` event. 

This got messed up in the commit - https://github.com/PolymeshAssociation/polymesh-subquery/pull/236/files#diff-45b7bc2074fd178435206ac01414cde32b1413b2e783a1c568a7ab6f4fbc14d0L362 where in refactoring the code to project.ts file and renaming of the method 

### Breaking Changes

Since the entries for `TransactionFeePaid` event where not indexed, this will require a full re-sync

### Checklist

### DIFF
--- a/project.ts
+++ b/project.ts
@@ -312,8 +312,7 @@ const filters = {
     NewAccount: [],
   },
   transactionPayment: {
-    TransactionFeePaid: [],
-    FeeCharged: ['handleTransactionFeeCharged'],
+    TransactionFeePaid: ['handleTransactionFeeCharged'],
   },
   treasury: {
     TreasuryDisbursement: ['handleTreasuryDisbursement'],


project.ts
```diff
@@ -312,8 +312,7 @@ const filters = {
     NewAccount: [],
   },
   transactionPayment: {
-    TransactionFeePaid: [],
-    FeeCharged: ['handleTransactionFeeCharged'],
+    TransactionFeePaid: ['handleTransactionFeeCharged'],
```
polymath-eric, line 315: Why not `FeeCharged` anymore?

prashantasdeveloper: FeeCharged was never there in transactionPayment. When you refactored the project.ts it got messed up[ here](https://github.com/PolymeshAssociation/polymesh-subquery/pull/236/files#diff-45b7bc2074fd178435206ac01414cde32b1413b2e783a1c568a7ab6f4fbc14d0L362)

PR #294, opened 2025-02-06
fix: 🐛 correctly process instructionId for Transferred balance updates

F-OBrien:
Fix `InstructionId` for `AssetBalanceUpdated` events with reason `Transferred`. `InstructionId` and `instructionMemo` were incorrectly being processed as Codec types but had already been converted to JSON.

### Breaking Changes

none - Resync required to correct data


src/mappings/entities/assets/mapAsset.ts
```diff
@@ -559,14 +559,12 @@ export const handleAssetBalanceUpdated = async (event: SubstrateEvent): Promise<
...
-      readonly instructionMemo: Option<U8aFixed>;
+      instructionId: number | null;
+      instructionMemo: `0x${string}` | null;
     };
 
-    instructionId = details.instructionId?.isSome
-      ? processInstructionId(details.instructionId.unwrap())
-      : null;
-    instructionMemo = bytesToString(details.instructionMemo);
+    instructionId = details.instructionId ? details.instructionId.toString() : '';
```
prashantasdeveloper, line 566: Instead of empty string we should set it to undefined/null. That way querying is easier

prashantasdeveloper (changes requested):

prashantasdeveloper (approved):

PR #295, opened 2025-02-06
fix: 🐛 set package version

polymath-eric:
### Description

set package version so dev instances return a mostly accurate version

### Checklist


polymath-eric:
I tried with lint-staged. Its meant for files already committed. When I tried to put in the script there it would run, but not commit package.json like the linted files.

PR #304, opened 2025-05-08
feat: 🎸 publish arm docker image

polymath-eric:
### Description

publish arm64 docker image in addition to amd64

The docker docs say its this easy: https://docs.docker.com/build/ci/github-actions/multi-platform/

### Checklist

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -5,7 +5,7 @@ on:
     tags:
       - v*
   release:
-    type: [published]
+    types: [published]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image.yml
```diff
@@ -5,7 +5,7 @@ on:
     tags:
       - v*
   release:
-    type: [published]
+    types: [published]
```
polymath-eric, line 8: seems like `type` was deprecated as a key

VictorVicente (approved):

PR #309, opened 2025-06-03
chore: 🤖 add test:integration action

polymath-eric:
add package.json script action

### Checklist


docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
```
polymath-eric, line 1: This was giving a warning about being deprecated with the newer compose

PR #337, opened 2026-07-08
ci: add ecr publish workflow and configurable workers

debelio:
### Description

Renames `docker-image.yml` → `build-publish-image.yml` and adds an ECR push (via OIDC role assumption) alongside the existing Docker Hub publish, so images are also available in our AWS  account.

Also adds a configurable `MAX_WORKERS` env var (`--workers`) to `docker-entrypoint.sh`, defaulting to `2`.

### Breaking Changes

None.

### JIRA Link

None.

### Checklist

None.


docker/docker-entrypoint.sh
```diff
@@ -26,9 +26,11 @@ npm run migrations
 # Allow configuring node memory. It should be no more than 75% of available RAM.
 # Defaults to 3 GB (expecting at least 4 GB available).
 NODE_SPACE=${MAX_OLD_SPACE_SIZE:-3072}
+NODE_WORKERS=${MAX_WORKERS:-2}
```
prashantasdeveloper: We already had option to pass workers [here](https://github.com/PolymeshAssociation/polymesh-subquery/blob/master/docker-compose.yml#L58). Do we need it in entrypoint script as well?

debelio: You are right. Having "@" will use whatever command args are supplied at run time. Reverting this.

prashantasdeveloper (approved):

PR #348, opened 2026-09-08
feat: add decode layer, anomaly recording and chain-upgrade tracking [redesign(03)]

prashantasdeveloper:
## Phase 3 — Infrastructure

PR 4 of 10 in the indexer redesign series. Based on `redesign/02-claims` (PR #346).

Ships no consumer-visible feature. This is the decode layer, anomaly recording, and the `ChainUpgrade` hook that Phases 4–6 build on — doing it first means the model rewrites in those phases fail loudly rather than silently.

### What's in this PR

This is the largest PR in the series (11 commits). Please review **commit by commit** rather than by the combined diff — each is scoped to one concern.

1. **`feat: 🎸 add the IndexerAnomaly entity`** — converts silent corruption into a queryable defect list. An empty table after a full resync is the acceptance signal.
2. **`feat: 🎸 persist ChainUpgrade and drop module-level upgrade state`** — `mapChainUpgrade.ts` held upgrade-detection state in module-level variables, unsafe under `--workers` (per-worker rather than per-chain). Also folds `mappingHandlers.ts`'s per-block dedup state (B9) into a block-scoped context object.
3. **`feat: 🎸 retire ChildIdentity rows at the v8 boundary`** (defect A11) — the v8 upgrade deletes every child identity in a storage migration that emits no events, so the indexer retained stale rows forever. Driven off the persisted `ChainUpgrade` crossing.
4. **`feat: 🎸 decode event fields by name`** — new `src/decode/` module. Struct-style events (Metadata v14+) resolve parameters by the field name the block's own metadata carries, immune to field insertion/reordering.
5. **`feat: 🎸 add the legacy tuple decoder table with arity assertions`** — positional decoding for pre-7.x tuple events, keyed by spec range, with `NoDecoderForSpecVersion`/`ArityMismatch` anomalies on mismatch. Also folds the `polymesh_private_dev` spec offsets into one normalisation.
6. **`refactor: 💡 migrate surviving handlers onto the decode layer`** — settlement, asset, identity and external agents. Deliberately excludes balances and staking, which Phase 4 rewrites natively on the decode layer.
7. **`test: 💍 add fixture and metadata-contract tests for decoded events`** — per-event fixture tests, plus a metadata-contract test asserting every registered decoder's declared arity matches the checked-in metadata's actual arity for each captured spec version.
8. **`refactor: 💡 replace getPaginatedData with store.getByFields`** — rework, not an introduction. Phase 1 (`714e1a7`) already fixed the ordering bug (A13); this replaces the hand-rolled helper with `store.getByFields` across all seven call sites, preserving `id`-ordering.
9. **`perf: ⚡ cache account resolution per block`** — per-block cache with negative caching for `getOrCreateAccount`, the hottest chain-read path (reached twice per asset movement on v8).
10. **`chore: 🤖 consolidate index declarations into schema.graphql`** — moves every index `db/compat.sql` could express as `@index`/`@compositeIndexes` into the schema; only expression indexes, generated JSONB columns and the JSONB path index remain in `compat.sql`, each commented. Also fixes `docker-entrypoint.sh`'s startup race (`compat.sql` ran backgrounded with a self-kill, racing the node's schema creation).
11. **`feat: 🎸 add scripts/sync-metadata.ts`** — regenerates the three chain enums from runtime metadata deterministically, and reports new/changed/removed events per spec version plus events that are registered but never handled (~150 sit as `[]` today). No migration generation (D5: full resync). Deletes the stale `spec_diffs/` (stopped at 5003000).

**If review stalls, there is a clean split point after commit 7**: commits 1–7 are the decode layer and anomaly plumbing that Phases 4–6 depend on; 8–11 are independent cleanups nothing blocks on.

### Consumer impact

**None.** `IndexerAnomaly` and `ChainUpgrade` are additive; index consolidation is transparent (the indexes already existed, they just move to being declared in one place); no existing entity changes shape.

### Verification

```
yarn codegen && yarn typecheck && yarn lint && yarn test:unit
```

All green — 359 tests across 26 suites. No `db/migrations/*` entries (decision D5, full resync from genesis).


scripts/sync-metadata.ts
```diff
@@ -0,0 +1,474 @@
...
+  const snapshot: RuntimeSnapshot = {
+    specName,
+    specVersion,
+    modules: [],
+    events: {},
+    calls: {},
+  };
+
+  for (const pallet of metadata.asLatest.pallets) {
+    const section = pallet.name.toString();
```
F-OBrien, line 84: `snapshot.events` and `snapshot.calls` end up keyed by the metadata spelling of the pallet (`Asset`, `ExternalAgents`), but every consumer of a section here uses the `event.section` spelling (`asset`, `externalAgents`) — that's what the arity fixtures, `CAPTURED_MODULES` and `project.ts` all use. The two never meet.

Consequences:

- `eventDrift` looks up `snapshot.events[moduleId]` with a fixture key, misses, and falls back to `{}` — so every event in the fixture is reported as `removed`. Against mainnet 8000020 that's `added (0) / removed (81) / reshaped (0)`: the fixture's entire contents, with `added` and `reshaped` structurally unreachable since both derive from the same empty `current`.
- `arityFixtureFor` filters `CAPTURED_MODULES` through the same lookup, so `--write` emits a fixture with `"modules": {}`. That also makes `decodeContract.test.ts` pass vacuously for that spec version, which is the check this fixture exists to drive.

The drift report is the part of this script that's meant to catch a reshaped event, and right now it cannot — a real arity change would be indistinguishable from the noise.

Worth noting the two reports disagree with each other in a single run: `unhandledEvents` lowercases the section before comparing, so it lists `asset.AllowanceSpent` as present in the runtime while `eventDrift` calls it removed, off the same `snapshot` object. The enum reports are unaffected — `planEnumUpdates` reads module names and event values, never section keys.

Suggested fix — normalise once on the way in, so the metadata spelling never leaves `snapshotFromMetadata`:

```ts
/** A pallet name as `@polkadot/api` reports `event.section`: `ExternalAgents` -> `externalAgents` */
const sectionId = (name: string): string => name[0].toLowerCase() + name.slice(1);

// in the pallet loop:
const section = sectionId(pallet.name.toString());
```

`snapshot.modules` is unaffected — `section.toLowerCase()` gives the same fully-lowercased value either way, which is what `ModuleIdEnum` wants.

With that applied, mainnet 8000020 against the 8000000 fixture goes to `added (0) / removed (0) / reshaped (0)`.

One note on the tests: `tests/unit/syncMetadata.test.ts` currently keys both the fixture and the snapshot `Balances`, so the suite is green over the broken script — the producer and consumer conventions are never crossed. Fixing that convention and adding a case built on real metadata (`@polkadot/types-support/metadata/static-substrate`, already a dependency, so no chain needed) covers it: capture a fixture from a runtime and read it back against that same runtime, and it must show no drift.

F-OBrien (approved):

PR #349, opened 2026-09-09
feat!: rebuild the POLYX record as a double-entry ledger (redesign phase 04)

prashantasdeveloper:
## Summary

Phase 4 of the indexer redesign — plan [`docs/implementation/02-polyx-ledger.md`](docs/implementation/02-polyx-ledger.md).
Replaces the single-column `PolyxTransaction` list with an entry-centric `PolyxEntry`
ledger plus a materialised `AccountBalance`, so "what was account X's balance at
block N" becomes an index scan instead of an un-answerable question.

Chained on `redesign/03-infrastructure` (decode layer + `IndexerAnomaly`).

## Why

`PolyxTransaction` had four structural faults, each fatal to point-in-time balances:

1. **One `type` column for a two-sided movement** — `Reserved` is `Free → Reserved`;
   a single row records half of it.
2. **`frozen` is a MAX of active locks, not a SUM** — no summing of movement rows
   reproduces it.
3. **`BalanceSet` is an absolute value recorded as a delta** — corrupts every running
   total after it.
4. **`BalanceTypeEnum` conflated three systems** — real pools (`Free`/`Reserved`), a
   lock floor, and staking-ledger states whose mechanism inverted at v8.

Plus: several v8 `balances` events were unregistered, leaving `reserved` unindexed on v8.

## What changed

**Model**
- **`PolyxEntry`** — append-only, one row per `(account, movement side)`, carrying
  `freeAfter` / `reservedAfter` / `frozenAfter` snapshots; sibling entries of one
  on-chain movement share `movementId`. New `MovementKind` / `EntryDirection` /
  `PolyxPool` / `HoldReason` enums replace `BalanceTypeEnum`.
- **`AccountBalance`** — running balance, height-versioned. `frozen = MAX(locks)`,
  `bonded` = staking lock (≤ v7.4) or `Staking` hold (v8), plus `transferable`,
  `otherReserved`, and lifetime aggregates.
- `PolyxTransaction` and `BalanceTypeEnum` removed.

**Handlers** — `mapPolyxLedger.ts` replaces `mapPolyxTransaction.ts`
- Every `balances.*` movement (transfer / endow / reserve / repatriate / hold /
  release / burn / mint / dust / `BalanceSet`) writes entries and advances the
  balance, at both the pre-v8 tuple shape and the v8 struct shape.
- Staking: pre-v8 `set_lock` bonding, `RewardDestination::Staked` auto-restake
  (credited *and* locked, no `Bonded` event), v8 `Bonded ↔ balances.Held{Staking}`.
- Treasury reimbursement / disbursement and transaction-fee attribution.

**Reconciliation**
- In-flight (`reconcilePolyx.ts`) — periodically checks the derived `AccountBalance`
  against `system.account` on chain, records a `BalanceReconciliationDrift` anomaly
  and corrects, so one missed event can't compound.
- Offline (`scripts/reconcile-polyx.ts`) — full-history audit.

**Decode shapes** — pre-v8 tuple shapes for `balances.*` and `staking.*`.

**Postgres 18** — the height-versioned indexes need `btree_gist`'s boolean operator
class (PG 13+); image `postgres:12-alpine → 18-alpine`, data mount adjusted for the
PG 18 subdirectory layout.

**Full resync from genesis** (decision D5) — `db/migrations/*` deleted; account
resolution fixed for reindex-from-zero.

## Follow-up commits

Three `chore` commits refine the new code (all pre-merge, none touch released code):

- `settlement.InstructionCreated` accepts its pre-`memo` 7-param arity.
- skip `identity.ClaimRevoked` with a zero issuer (stripped early-chain event).
- treasury-reimbursement crediting, pre-v8 `Staked` restake lock, `staking.payee()`
  `.toJSON()` variant normalisation, a 100-POLYX reconciler noise floor; plus a
  per-stash payee cache, a coarser reconcile interval and per-block `system.account`
  memoisation.

## Breaking changes

- Schema: `PolyxTransaction`, `BalanceTypeEnum` removed; `PolyxEntry`,
  `AccountBalance`, `PolyxPool`, `MovementKind`, `EntryDirection`, `HoldReason` added.
- Postgres 12 → 18 (image + volume path).
- Requires a full resync from genesis.

## Testing

- Gate green: `yarn codegen && yarn typecheck && yarn lint && yarn test:unit`
  (396 unit tests; new suites `mapPolyxLedger`, `reconcilePolyx`, `seedAccountBalances`,
  `rewardDestinationA15`, `extract8xStakingAmount`).
- Validated by a full resync from genesis; the issues it surfaced are the three
  follow-up commits above.

## Known follow-up

The pre-v8 staking lock is derived by accumulating restaked rewards; on chain it is
`staking.ledger.active` (max-bond capped, and MAX'd against other locks). The in-flight
reconciler bounds the resulting drift to ≤ ~1 era per validator and corrects it at each
checkpoint; a precise fix (read `ledger.active` at the reward block) is tracked for a
follow-up.

## Merge

Rebase-merge onto `redesign/03-infrastructure`. Do not squash.


src/utils/accounts.ts
```diff
@@ -36,6 +44,47 @@ export const getAccountKeyType = (
...
+ * `identity.keyRecords` is a 5.x rename of `identity.keyToIdentityIds`; a genesis resync sees the
+ * older name on early blocks. The legacy storage is `Option<IdentityId>` on both public chains
+ * (Polymesh launched at v3, so there is no `LinkedKeyInfo` enum to unwrap), and primary vs
+ * secondary is read from `identity.didRecords`.
+ */
+const resolveKeyIdentity = async (
+  address: string
+): Promise<{ did: string; type: 'primaryKey' | 'secondaryKey' } | undefined> => {
+  if (typeof api.query.identity.keyRecords === 'function') {
+    const raw = (await api.query.identity.keyRecords(address)) as unknown as Codec;
```
F-OBrien: `as unknown as Codec` is unnecessary here and could mask a genuine type issue. remove the casting

src/utils/accounts.ts
```diff
@@ -36,6 +44,47 @@ export const getAccountKeyType = (
...
+      did: getFirstValueFromJson(raw),
+      type: getFirstKeyFromJson(raw) === 'primaryKey' ? 'primaryKey' : 'secondaryKey',
+    };
+  }
+
+  const raw = (await legacyQuery(
+    'identity',
+    'keyToIdentityIds',
+    [3000, 5_000_002]
+  )(address)) as unknown as Codec;
```
F-OBrien: `as unknown as Codec` casting is unnecessary here. The query already returns a `Codec` type

src/utils/accounts.ts
F-OBrien: This can be ignored the next PR resolves this

src/utils/staking.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Per-stash cache of a resolved payee. `staking.payee(stash)` is a chain-storage read on every
+ * reward, and a validator's set of ~20 stashes is re-queried every era for the whole genesis
+ * replay — the dominant RPC cost of the sweep against a remote node. A payee changes very rarely
+ * (an explicit `staking.setPayee`), and the in-flight reconciler corrects any `frozen` drift a
+ * stale entry could cause within ~1 era, so a plain process-lifetime cache is the right trade.
+ * Only successful resolutions are cached — a transient read failure must be retried, not pinned.
+ */
+const payeeCache = new Map<string, LegacyRewardDestination>();
```
F-OBrien: The staking caches never pick up payee or controller changes

`payeeCache` and `controllerCache` are module-level Maps keyed only by stash. They live for the whole indexer process, and nothing clears an entry after it's written. If a stash later calls setPayee or setController, every reward and lock read after that still uses the value cached the first time.

The comment says the in-flight reconciler corrects any frozen drift within about one era. That covers the lock balance, but not everything these caches feed:

- Reward records keep the wrong value. `rewardDestination` / `rewardDestinationAccount` from `resolveLegacyRewardDestination` are saved on the reward entities, and nothing reconciles those afterwards. A stale payee leaves a permanently wrong record of where the POLYX went. A cached `Controller` result has the same problem, because it stores the controller as it was at first read.
- The `bonded` fallback gets cached. In `readStakingLock`, a missing `bonded(stash)` falls back to `controller = stash`, and that value is cached (staking.ts:170-171). This goes against the "only successful resolutions are cached" rule used in the payee path. If this runs on an unbonded stash that later bonds with a different controller, `ledger(stash)` would be read from then on and the lock would come back as `0`.
- Results depend on the process. Each `--workers` thread and each restart starts with an empty cache. The same block can produce different reward destinations depending on where the sync resumed.

I understand the cache is to cut the RPC cost of the genesis replay but it's not ideal in the current form. I understand there's more staking indexing work coming, and I think this is worth picking up there again. Some options:

- Clear a stash's entries when handling `setPayee` / `setController` (and `bond`/`unbond` for the controller mapping).
- Don't cache the bonded fallback, only a controller actually read from chain.
- If invalidation turns out to be impractical, key by era, but then we are probably negating most of the benefit as the cache will only serve stashes with multiple payouts per era.

src/mappings/entities/assets/mapNfts.ts
```diff
@@ -14,13 +14,56 @@ import {
...
+
 export const getNftHolder = async (
   assetId: string,
   did: string,
   blockId: string
 ): Promise<NftHolder> => {
   const id = `${assetId}/${did}`;
 
+  // A holder mutated earlier in this same block lives in the buffer, not yet in the store.
+  const buffered = bufferedBlock === blockId ? bufferedHolders.get(id) : undefined;
```
F-OBrien: Bug: a holder changed in two blocks within one flush interval loses the first block's changes

`getNftHolder` only returns the buffered copy when it belongs to the *current* block. `bufferHolder`, which is what flushes the previous block, runs *after* `getNftHolder`. Walking through one holder H:

1. Block N: `issued` NFT 5 to H. H is buffered with `[…, 5]` and not saved.
2. Block M, before the next `handleBlock` flush (height % 100): `issued` NFT 6 to H.
   - `getNftHolder(H, M)`: `bufferedBlock` is N, so it falls through to `NftHolder.get` and gets H **without 5**.
   - It pushes 6, then `bufferHolder` flushes N's copy (with 5) and buffers M's copy (without 5).
3. The next flush saves M's copy, and **NFT 5 is lost from H**.

`transferred` and `redeemed` are affected the same way. For example, a transfer out in block M can put back an NFT that was removed in block N. Any holder touched twice within 100 blocks is affected, so active collections will hit this routinely.

The doc comment says holders are "saved once, when the block changes". In practice the save happens when an NFT event from a *later* block arrives, or at the next 100-block `handleBlock`, not at the block boundary.

Suggested fix: flush a stale buffer before reading from the store, once per event and not per `getNftHolder` call (the transfer path calls it twice concurrently):

```ts
// at the top of handleNftHoldingsUpdates
if (bufferedBlock !== undefined && bufferedBlock !== blockId) {
  await flushNftBuffer();
}
```

A related point to consider: even with this fix, block N's holders are saved while block M is being processed. In historical mode I'd expect the versioned row to start at M rather than N, while `updatedBlockId` says N. A test that mints to the same holder in two blocks, with no flush in between, would cover the lost update.


Changes in the buffer could also be lost on a restart

PR #351, opened 2026-09-09
fix: 🐛 read key records through the augmented types

F-OBrien:
Found while reviewing #349. Draft — #349 is still under review.

### The bug

`identity.keyRecords` was cast to a bare `Codec` and read via `.toJSON()`, which hid that `KeyRecord` has **three** variants. `MultiSigSignerKey` carries the **multisig's account**, not a DID, so a signer key produced:

- an `Identity` row keyed by an SS58 address
- an `Account` pointing at that fabricated identity
- a portfolio 0 for it

A signer key has no identity and no permissions, so it now resolves to no `Account` at all. `ledgerAccount` still creates a bare one if the address ever holds POLYX, which is the same row it creates for a pallet or system address. No DID is resolved for a signer on purpose: the signer→multisig and multisig→identity links rebind independently, and `multiSig.adminDid` is the admin, not the identity the multisig is joined to.

`MultiSigSigner` rows are left to the multisig event handlers and the genesis/seed scan — neither `status` nor `createdBlock` is derivable from a key record.

### The cache

`BlockContext.accounts` held the resolved `Account`, so a cache hit skipped the **store** read as well as the chain read. The chain read cannot change within a block (`api` is `api.at(blockHash)` of the block being indexed, so it serves end-of-block state), but the row can — a handler links or unlinks a key partway through. So a cached negative shadowed a row another handler had just written, and `ledgerAccount` then overwrote it, dropping its `identityId`.

It now caches only the key record. `Account` is read from the store on every lookup. Two regression tests cover both staleness modes; they fail if entity caching is put back.

### Casts

Removed where they were discarding real augmented types: both `system.account` reads and `scripts/reconcile-polyx.ts`, narrowed to the balance fields — the only part that must stay spec-agnostic, since `frozen` is `miscFrozen`/`feeFrozen` on older runtimes.

Kept where they are load-bearing: every `event.event.data[i]` / `extrinsic.args[0]` site. Those pass `tsc` and fail `subql build` — two physical copies of `@polkadot/types-codec` (`cjs/types/codec` vs `types/codec`) that the two toolchains resolve differently. They mask nothing: event data is `Codec[]`, so there are no augmented types on that path.

`scripts/reconcile-polyx.ts` now loads the chain-type augmentation itself. Only `tsconfig.test.json` includes `scripts/**/*`, so an editor was type-checking it against a different, unaugmented type than CI — worth closing regardless of this PR. The other two chain-reading scripts still have that gap.

### Verified

`tsc --noEmit -p tsconfig.test.json` clean, `eslint src scripts tests` clean, `subql build` exits 0, 409/409 unit tests pass.

### Not included

Pre-existing rows in any already-synced database still have the fabricated identities; that needs a reindex or migration.

### DIFF
--- a/scripts/reconcile-polyx.ts
+++ b/scripts/reconcile-polyx.ts
@@ -29,6 +29,13 @@
  *       --rpc wss://mainnet-rpc.polymesh.network \
  *       [--sample 80] [--high 30] [--typical 30] [--reset]
  */
+// Chain-type augmentation, in the order `src/index.ts` loads it: `types-lookup` supplies the
+// types `augment-api` imports, and `@polkadot/api-augment` is deliberately not loaded alongside
+// (docs/implementation/12-types-and-ci.md §12.2). `tsconfig.test.json` already pulls this in via
+// `src/**/*`, but a script is also run and edited on its own, so it declares what it depends on.
+import '@polkadot/types-augment';
+import '@polymeshassociation/polymesh-types/polkadot/types-lookup';
+import '@polymeshassociation/polymesh-types/polkadot/augment-api';
 import { ApiPromise, WsProvider } from '@polkadot/api';
 import { existsSync, readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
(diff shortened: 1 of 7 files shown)


prashantasdeveloper:
I have added this commit in https://github.com/PolymeshAssociation/polymesh-subquery/pull/353#top

PR #353, opened 2026-09-10
feat!: 🎸 rework the identity/key and multisig model (redesign 6/10)

prashantasdeveloper:
# Identity & keys

Implements `docs/implementation/04-identity-keys.md`. Models key membership as a time-bounded
record, links the multisig cluster to `Account`, adds a key-role field, and drops dead columns.

Based on `redesign/05-holdings-nfts`. Parallel track with phases 5 and 8; needs a rebase before merge.

## Schema

**Added**

- `IdentityKey` — one row per `(identity, account, role)` membership interval. `validFromBlock` /
  `validToBlock` (null = active), `addedReason` / `removedReason`, `permissions: PermissionsJson`.
  Composite indexes `[identity, role]` and `[account, validToBlock]`.
- `enum KeyRole { Primary Secondary }`.
- `Identity.keys: [IdentityKey!]! @derivedFrom`, `Account.keyAssignments: [IdentityKey!]! @derivedFrom`.
- `Account.keyRole: KeyRoleEnum!` (`PrimaryKey` / `SecondaryKey` / `MultiSigSigner` / `Unlinked`), indexed.
- `MultiSigSigner.signerAccount: Account` (nullable), indexed.
- `Account.multiSig: MultiSig @derivedFrom`.
- `AuthorizationStatusEnum.RetryLimitReached`.

**Changed**

- `MultiSig.address: String!` → `account: Account!`.
- `MultiSigAdmin.identityId: String!` → `admin: Identity!`.
- `MultiSig.creator` / `creatorAccount` → nullable; populated only from `MultiSigCreated`.

**Removed**

- `Identity.secondaryAccounts` (derived field; wrong semantics, `@derivedFrom` cannot filter).
- `AccountHistory` entity.
- `Permissions` entity and `Account.permissions`.
- `Event.claimType` / `claimScope` / `claimIssuer` / `claimExpiry` / `corporateActionTicker` /
  `fundraiserOfferingAsset` / `transferTo`, plus their `db/compat.sql` indexes.

## Behaviour

- Identity handlers (`handleDidCreated`, `handleSecondaryKeysAdded`,
  `handleSecondaryKeysPermissionsUpdated`, `handleSecondaryKeysRemoved`, `handleSignerLeft`,
  `handleSecondaryKeyLeftIdentity`, `handlePrimaryKeyUpdated`), `genesisHandler`, and the lazy
  `getOrCreateAccount` path open/close/rotate `IdentityKey` rows and set `keyRole`.
- Pre-5.0 `identity` payload unwrapping moved from inline handler duck-typing to `src/decode/legacy.ts`.
- Multisig handlers create the multisig's `Account`, link `MultiSig.account`, populate
  `MultiSigSigner.signerAccount` (Account signers only), and set `keyRole: MultiSigSigner` on
  signer accounts.
- `genesisHandler.handleMultiSigs` seeds `MultiSigAdmin` from `multiSig.adminDid` on all chain
  versions; leaves `MultiSig.creator` null.
- `AuthorizationRetryLimitReached` marks the `Authorization` row `RetryLimitReached` via
  `handleAuthorization`; the status handler tolerates a missing row.
- `ledgerAccount` moved from `mapPolyxLedger.ts` to `src/utils/accounts.ts`.

## Decisions

- **`MultiSigSigner.signerAccount` is nullable.** `SignerTypeEnum` is `Account | Identity` and
  pre-7.x runtimes allowed identity signers; a relation cannot point at those. `signerValue`
  stays the canonical value.
- **`KeyRoleEnum.Unlinked` is one value.** The chain's `KeyRecord` is a closed three-variant enum
  and every other address is `None`; pallet addresses, system pots and detached keys are
  indistinguishable from chain state. The enum is left open for a later indexer-side split.
- **`MultiSig.creator` split from admin.** The chain models creator (`MultiSigCreated.callerDid`),
  admin (`multiSig.adminDid`), paying identity (`multiSig.payingDid`) and joined identity
  (`identity.keyRecords`) as four separate relationships. `creator` is now creator-only and
  nullable; admin is `MultiSig.admins`; joined identity is `Account.identity`.
- **`CddClaimsInvalidated` left unhandled.** It invalidates CDD claims, not keys, and is absent
  from the v8 runtime — belongs with plan 01 (claims).
- **`Event` denormalised columns dropped, not consolidated** (per `09-infrastructure.md §9.4`).
  Harvester carry-over, unqueried by either consumer, duplicated by `Claim` / corporate-action /
  STO entities.

## Not in this PR

- G2 — `Identity.primaryAccount` stays a string (use `keys(filter: { role: Primary })`).
- Multisig handlers still use positional decoding — moving to `decodeEvent` needs
  `src/decode/shapes/multiSig.ts`.
- `multiSig.payingDid` is unindexed (`MultiSigRemovedPayingDid: []`, no "set" event).

## Consumer impact

`AccountHistory`, `Permissions`, `MultiSig`, `MultiSigAdmin`, `MultiSigSigner`, `ChildIdentity` are
not directly queried by the SDK or portal (`consumer-queries.md`). `Identity.secondaryAccounts` is
read from chain by the SDK, not middleware. The removed `Event` columns are neither selected nor
filtered by either consumer. `authorizations` is queried; a new enum value is additive.

## Migrations

None — `db/migrations/*` is not used (D5, full resync from genesis). `tests/entities/*` snapshots
regenerate on that resync.

## Verification

`yarn codegen && yarn typecheck && yarn lint && yarn test:unit` — pass (42 suites, 482 tests).
`yarn check-handlers` — pass.


schema.graphql
```diff
@@ -2232,6 +2265,8 @@ enum AuthorizationStatusEnum {
   Consumed
   Rejected
   Revoked
+  "The chain gave up re-offering the authorization after too many failed accept attempts"
+  RetryLimitReached
```
F-OBrien, line 2269: the `AuthorizationRetryLimitReached` event is declared but never emitted. Harmless having it here. We can infer the status if the count on an authorization goes to 0 but the count isn't emitted in an event so I expect they will just sit at Pending.

src/mappings/entities/identities/mapIdentities.ts
```diff
@@ -60,18 +46,6 @@ const getIdentity = async (did: string): Promise<Identity> => {
...
-  blockId: string
-): Promise<void> =>
-  Permissions.create({
-    id: address,
-    ...args,
-    createdBlockId: blockId,
-    updatedBlockId: blockId,
-  }).save();
-
 export const createAccount = async (
```
F-OBrien, line 49: This will override and existing account which will clobber the original createdBlock. Should we consider an updateOrCreate option and should this be in the account util file?

src/mappings/entities/identities/mapIdentities.ts
```diff
@@ -312,77 +284,55 @@ const getPermissions = (accountPermissions: Record<string, unknown>): Permission
...
-      ({ account: address } = account);
-    }
-
-    return [Account.remove(address), Permissions.remove(address)];
-  });
-
-  await Promise.all(removePromises.flat());
+  await Promise.all(
+    addresses.flatMap(address => [
+      Account.remove(address),
```
F-OBrien, line 320: `Account.remove(address)` deletes the row that this PR's new `IdentityKey` history points at. `closeIdentityKeys` sets `validToBlock`/`removedReason` on the key's open interval (`accountId = address`) while `Account.remove` runs concurrently in the same `Promise.all`, so the closed interval references a row that no longer exists — and `IdentityKey.account` is a non-null relation. `account { address }` on a historical interval resolves to nothing, `Account.keyAssignments` becomes unreachable, and any `PolyxEntry.account!` / `AccountBalance.account!` rows point at the same deleted account.

`handleSignerLeft` (line 333) does the same thing. `handleSecondaryKeyLeftIdentity` (line 465) models the identical transition the other way — keep the row, clear `identityId`, set `keyRole = Unlinked` — which looks like the behaviour we want in all three now that leaving is recorded on `IdentityKey` rather than by deleting the account.

---

Worth adding: this path is hit during every primary-key rotation, not just a plain removal. On [24971813-1](https://polymesh.subscan.io/extrinsic/24971813-1) and [21528057-1](https://polymesh.subscan.io/extrinsic/21528057-1) the chain emits `SecondaryKeysRemoved(did, [new_primary])` immediately *before* `PrimaryKeyUpdated`, so `Account.remove` here deletes the account that is about to become the identity's primary key. It survives only because `handlePrimaryKeyUpdated` recreates it two events later — and that recreate resets its `createdBlock` (see line 49). Keeping the row and setting `keyRole = Unlinked` would make both the removal and the rotation case behave.

src/mappings/entities/identities/mapIdentities.ts
```diff
@@ -468,89 +407,71 @@ export const handleSecondaryKeysAdded = async (event: SubstrateEvent): Promise<v
...
 
   const did = getTextValue(rawDid);
   const address = getTextValue(rawNewKey);
 
   const identity = await getIdentity(did);
-  const [account, permissions] = await Promise.all([
-    Account.get(identity.primaryAccount),
-    Permissions.get(identity.primaryAccount),
-  ]);
+  const account = await Account.get(identity.primaryAccount);
```
F-OBrien, line 418: `Account.get` returns `Account | undefined` and the result is used unguarded on the lines below. `createIdentityIfNotExists` writes `primaryAccount: ''` (line 87), so an identity created down that path that later emits `PrimaryKeyUpdated` gives `Account.get('')` → `undefined` → `TypeError`, killing the block rather than skipping the update.

Same unguarded read at line 465 in `handleSecondaryKeyLeftIdentity`. Narrow path in both cases, but a guard plus `recordAnomaly` beats a crash.

src/mappings/entities/identities/mapIdentities.ts
```diff
@@ -468,89 +407,71 @@ export const handleSecondaryKeysAdded = async (event: SubstrateEvent): Promise<v
...
+    // close the old primary's membership interval — the rotation history lives on `IdentityKey`
+    closeIdentityKeys(
+      { address: account.id, role: KeyRole.Primary, removedReason: eventId },
+      blockId
+    ),
   ]);
+
+  // ...and open the new primary's. The rotation record G3 asks for: both rows stay queryable, the
+  // old row's `validToBlock` equals the new one's `validFromBlock`.
+  await openIdentityKey(
```
F-OBrien, line 452: Withdrawing my earlier concern here — I checked two real rotations on mainnet and the Secondary interval *is* closed, just not by this handler.

`identity.acceptPrimaryKey` ([24971813-1](https://polymesh.subscan.io/extrinsic/24971813-1)) and `identity.rotatePrimaryKeyToSecondary` ([21528057-1](https://polymesh.subscan.io/extrinsic/21528057-1)) both promote an *existing secondary key*, and the chain emits the transition as a fixed sequence:

```
SecondaryKeysRemoved(did, [new_primary])   ← closes the incoming key's Secondary interval
PrimaryKeyUpdated(did, old, new)           ← this handler
SecondaryKeysAdded(did, [old_primary])     ← rotatePrimaryKeyToSecondary only
```

Handlers run in event order, so `handleSecondaryKeysRemoved` closes the Secondary interval before we get here and `handleSecondaryKeysAdded` re-links the demoted old primary after. No double-open, and the four-step rotation comes out right end to end.

Two things that follow, though:

1. **This handler is correct only because of its siblings, and nothing says so.** Read on its own it looks like the incoming key's Secondary interval leaks — that's what I assumed. Since this PR introduces the `IdentityKey` model, the three-event sequence is worth a docstring on `handlePrimaryKeyUpdated`, so the next reader doesn't "fix" the non-bug or reorder something that breaks it.

2. **The coupling is load-bearing in a fragile way.** At the first event, `handleSecondaryKeysRemoved` calls `Account.remove(new_primary)` — it deletes the account that is about to become the identity's primary key. It survives only because `createAccount` recreates it two events later, which also resets its `createdBlock` on every rotation (see line 49). That's a second argument for the `keyRole = Unlinked` treatment over `Account.remove` (line 320).

src/mappings/entities/identities/mapIdentityKey.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+export const rotateIdentityKey = async (
+  { address, role, reason, eventIdx, permissions, identityId }: RotateArgs,
+  blockId: string
+): Promise<void> => {
+  const [closed] = await closeIdentityKeys({ address, role, removedReason: reason }, blockId);
+
+  const owningIdentity = identityId ?? closed?.identityId;
+
+  if (!owningIdentity) {
```
F-OBrien, line 115: This bails *after* `closeIdentityKeys` has already committed. When the account has no open interval, `closed` is undefined and no `identityId` was passed — which is exactly the call shape at `mapIdentities.ts:300` for `SecondaryKeysPermissionsUpdated` — so we return having closed whatever was open, left the key with no active membership, and recorded the new permissions nowhere. Nothing is logged, so it's silent; the code this replaced threw here. `recordAnomaly` seems like the right call instead of a bare return.

Separately, `const [closed] = …` takes only the first row while `closeIdentityKeys` closes *all* open intervals for the account, so a key with more than one open interval has the rest closed and never reopened.

src/mappings/entities/multiSig/mapMultiSig.ts
```diff
@@ -15,52 +16,97 @@ import {
...
+  blockId: string,
+  datetime: Date
+): Promise<string | undefined> => {
+  if (signerType !== SignerTypeEnum.Account) {
+    return undefined;
+  }
+
+  const account = await ledgerAccount(signerValue, blockId, datetime);
+
+  if (account.keyRole !== KeyRoleEnum.MultiSigSigner) {
```
F-OBrien, line 70: This force-overwrites `keyRole` on a key that may currently be live in another role. `MultiSigCreated` / `MultiSigSignerAuthorized` are pending *offers* — the chain only writes `MultiSigSignerKey` when the signer accepts — so an account that is right now an identity's primary or secondary key gets `keyRole = MultiSigSigner` while its `identityId` and its open `role: Primary` `IdentityKey` row still say otherwise. `accounts(filter: { keyRole: { equalTo: PrimaryKey } })` then loses it.

Nothing reverts this if the offer is rejected or expires either. Deriving the role from `resolveKeyRole` (which reads `identity.keyRecords`, i.e. acceptance) rather than from the offer event would keep it honest — and would give `resolveKeyRole` its first caller.

src/mappings/entities/multiSig/mapMultiSig.ts
```diff
@@ -15,52 +16,97 @@ import {
...
+
+/**
+ * The `Account` id for `MultiSigSigner.signerAccount`, creating the account row if it is new.
+ *
+ * Only `Account` signers get one — `SignerTypeEnum` also has `Identity` on pre-7.x runtimes,
+ * which a relation cannot point at, so `signerValue` stays canonical and this returns undefined
+ * for those. A signer key is an account with no identity of its own, so `ledgerAccount` gives it
+ * a bare row; `keyRole` is set to `MultiSigSigner` from here because the event is proof of the
+ * role even before the chain writes the key record (which happens when the signer accepts). The
+ * genesis/seed scan later confirms it authoritatively from `multiSig.multiSigSigners`.
```
F-OBrien, line 56: Two claims in this paragraph don't hold:

- "The genesis/seed scan later confirms it authoritatively from `multiSig.multiSigSigners`" — nothing in `src/seed/` or `genesisHandler` writes `Account.keyRole` for a signer. The genesis path calls `createMultiSigSigner`, which reaches `linkSignerAccount` here, so it's the same write, not an independent confirmation.
- Nothing ever resets the role. `handleMultiSigSignerRemoved` / `handleMultiSigSignersRemoved` (→ `handleMultiSigSignerStatus`, line 141) only flip `MultiSigSigner.status`; the account stays `MultiSigSigner` forever. That contradicts the schema's "mutable, maintained on every key-link/unlink write" on `Account.keyRole`.

src/mappings/migrations/genesisHandler.ts
```diff
@@ -191,15 +193,16 @@ const handleMultiSigs = async (): Promise<void> => {
...
-        genesisBlock
+        genesisBlock,
+        datetime
       )
     );
 
-    if (is7xChainAtGenesis) {
-      createMultiSigAdmin(multiSigAddress, creator, genesisBlock);
+    if (adminDid.length) {
+      multiSigInserts.push(createMultiSigAdmin(multiSigAddress, adminDid, genesisBlock));
```
F-OBrien, line 205: `MultiSigAdmin.admin` is a relation now, but `adminDid` comes straight from chain storage unchecked. `handleGenesisDids` seeds only the 33 GC DIDs plus the systematic issuers, so a genesis multisig whose `adminDid` (or `multiSigToIdentity`) names any other DID writes an `admin_id` pointing at an `identities` row that doesn't exist.

To be precise about the failure mode: this project runs with historical tracking on, and under historical mode SubQuery does **not** emit foreign-key DDL. `createRelation` only registers the relation in `foreignKeyMap`, which becomes Postgres comments / smart tags for the GraphQL layer (`@subql/node-core`, `db/migration-service/migration.js`, the `if (this.historical)` branch); it can't emit a real constraint because `id` stops being unique once rows are versioned by `_block_range`. Nothing in `db/` adds one back. So nothing fails at insert time — the breakage lands at query time, where `admin` is declared `Identity!`, resolves to null, and the non-null violation errors the field and bubbles up through its parent in the response.

The old `identityId: String!` returned the dangling DID as a plain string, so consumers got a value that just didn't join to anything. The relation is still the right modelling call — it converts a silent dangling reference into a hard query error for anyone selecting `admin`, which is why it wants a guard rather than being left to chance.

Same gap at runtime in `handleMultiSigAddedAdmin` (`mapMultiSig.ts:203`) and in the pre-7.x `creator` path in `handleMultiSigCreated` (`mapMultiSig.ts:197`) — both hand a DID to `createMultiSigAdmin` with no existence check.

This gets much more likely once `START_BLOCK > 1` is a supported mode (`docs/implementation/10-partial-index.md`): `handleGenesis` is skipped entirely for a later start block, so *no* identities are seeded and every `MultiSigAddedAdmin` for a multisig predating the start block names a DID the index has never seen. What's an edge case at genesis becomes the default under partial indexing.

On the fix, I'd point at **Rule 2** of plan 10 rather than `createIdentityIfNotExists`: resolve from the DB, fall back to chain state at the current block, record an `IndexerAnomaly` on a second miss. `createIdentityIfNotExists` writes `primaryAccount: ''`, which is precisely the zero-filled row that rule exists to prevent.

src/utils/accounts.ts
```diff
@@ -82,47 +96,110 @@ const resolveKeyIdentity = async (
...
+};
+
+/**
+ * The `Account.keyRole` the chain's key record gives an address, read at most once per block.
+ *
+ * The one derivation path for the field: identity and multisig handlers, `getOrCreateAccount`,
+ * `ledgerAccount`, and the genesis/seed scan all resolve `keyRole` through this or `keyRoleFor`,
+ * so a role is never accumulated from events and cannot go stale relative to `keyRecords`.
+ */
+export const resolveKeyRole = async (address: string, blockId: string): Promise<KeyRoleEnum> =>
```
F-OBrien, line 160: `resolveKeyRole` has no callers, and the invariant this comment describes isn't implemented. Every writer hardcodes the enum instead — `mapIdentities.ts:162`, `:385`, `genesisHandler.ts:101`, `mapMultiSig.ts:71` — and only `ledgerAccount`'s bare-row path consults the chain, via `keyRoleFor`. So `keyRole` *is* accumulated from events and can go stale relative to `keyRecords`, which is what the two multisig `keyRole` comments come down to.

Either route the writers through `resolveKeyRole` and keep the guarantee, or drop the unused export and soften the comment to what the code actually does.

src/mappings/entities/multiSig/mapMultiSig.ts
```diff
@@ -120,20 +166,29 @@ export const handleMultiSigCreated = async (event: SubstrateEvent): Promise<void
...
+            signerValue,
+            blockId,
+            block.timestamp
+          ),
+          status: MultiSigSignerStatusEnum.Authorized,
+          createdBlockId: blockId,
+          updatedBlockId: blockId,
+        } satisfies MultiSigSignerProps)
+    )
   );
```
F-OBrien, line 192: The signers haven't accepted yet so should'nt be linked fully. Particularly if they can't accept (i.e. joined to another identity)

schema.graphql
```diff
@@ -2106,6 +2089,56 @@ type Identity @entity {
...
+  keys: [IdentityKey!]! @derivedFrom(field: "identity")
+}
+
+"""
+The role a key holds inside an Identity.
+
+Not a closed set — treat it as open. `MultiSigSigner` and detached-key roles are modelled on
+`Account.keyRole` (a separate enum) rather than here, and further membership roles may be added.
+"""
+enum KeyRole {
```
F-OBrien, line 2108: `KeyRole` and `KeyRoleEnum` are distinct enough to keep as two types — `MultiSigSigner` and `Unlinked` can't be values here (an "Unlinked membership interval" is a contradiction, and a signer key has no membership to hold a role in), so merging them would make invalid states representable. But the pair as named is confusing:

- **`Enum` carries no information.** Every legacy enum in this schema has the suffix (`EventIdEnum`, `ModuleIdEnum`, `SignerTypeEnum`), so `KeyRole` vs `KeyRoleEnum` is distinguished only by a token that means nothing. Neither name says which is scoped to a membership interval and which is the account's current state.
- **The value spellings diverge for the same concept** — `PrimaryKey`/`SecondaryKey` vs `Primary`/`Secondary`. GraphQL validation catches a mix-up so it isn't silent, but it's a papercut on every query and part of why `keyRoleFor` exists as a mapping.

**Recommendation:** rename both and align the spellings, so the subset relationship is visible on sight.

```graphql
"The key's current role per the chain's KeyRecord. Superset of IdentityKeyRole."
enum AccountKeyRole {
  PrimaryKey
  SecondaryKey
  MultiSigSigner
  Unlinked
}

"The role a key holds within one IdentityKey membership interval."
enum IdentityKeyRole {
  PrimaryKey
  SecondaryKey
}
```

with `Account.keyRole: AccountKeyRole!` and `IdentityKey.role: IdentityKeyRole!`.

Left as prose rather than a `suggestion` block because the change spans both enum declarations plus both field references — a partial apply would leave the schema invalid.

Cheapest to do now: both enums are new in this PR, so nothing downstream depends on the names yet. I checked `redesign/11-holdings-review` and both are still exactly as written here.

schema.graphql
```diff
@@ -2053,18 +2045,24 @@ type Account @entity {
   address: String! @index(unique: true)
   "`substrate` for a regular key, `ethereum` for the `0xEE` padded account of an Ethereum key"
   keyType: String!
+  "The key's role in the identity system, from the chain's `KeyRecord` — mutable, maintained on every key-link/unlink write"
+  keyRole: KeyRoleEnum! @index(unique: false)
```
F-OBrien, line 2049: `Account.keyRole` is the fourth representation of key membership in the schema, alongside `IdentityKey` (authoritative), `Account.identity`, and `Identity.primaryAccount`. Three of the four are hand-maintained caches of the first, and this review found a defect in each: `keyRole` goes stale for multisig signers and is never reset, `Account.identity` is cleared and then restored across two events during a rotation, and `primaryAccount` holds `''` as a sentinel that crashes `handlePrimaryKeyUpdated`.

G2 defers `primaryAccount` deliberately (`04-identity-keys.md:16`), which is fine. The gap is that the schema declares only `IdentityKey` to be the source of truth and says nothing about the standing of the other three.

**Recommendation:** a one-line docstring on each cache naming its derivation — here, something like "Derived from the chain's `KeyRecord`; the authoritative membership history is `keyAssignments`." That makes the staleness class findable by reading the schema, and sets up the G2 follow-up.

schema.graphql
```diff
@@ -2106,6 +2089,56 @@ type Identity @entity {
...
+A key joining an identity opens a row (`validFromBlock` set, `validToBlock` null). A key leaving,
+being rotated out, or having its permissions changed closes the row (`validToBlock`,
+`removedReason`) and, for a permissions change or rotation, opens a fresh one. The currently active
+membership for an account is the row with `validToBlockId: { isNull: true }` (`Boolean` cannot be
+indexed on the Postgres in use, so the null-FK filter is the query form).
+
+Replaces `Identity.primaryAccount`/`secondaryAccounts` as the source of truth for key membership,
+and subsumes `AccountHistory`, which had untyped columns and no validity interval.
+"""
+type IdentityKey @entity
```
F-OBrien, line 2125: Two consistency nits on the new entity:

- **The id format isn't documented.** `PolyxEntry` documents its own (`padId(block)/padId(eventIdx)/side`); this one is `identity/address/blockId/paddedEventIdx`. Consumers that page or construct ids need it.
- **No `datetime`**, where nearly every sibling entity carries one. Reachable via `validFromBlock`, so it's defensible — just inconsistent.

**Recommendation:** add the id format to the docstring; leave `datetime` off and rely on the `validFromBlock` join unless a query in `consumer-queries.md` actually needs date filtering without a block join.

schema.graphql
```diff
@@ -2073,20 +2071,6 @@ type PermissionsJson @jsonField {
   transactionGroups: [String]!
```
F-OBrien, line 2071: `transactionGroups` was a column on the `Permissions` *entity* and is now inside a jsonField, which SubQuery doesn't generate filters for. The rest of `Permissions` was already jsonFields, so this is the only field whose queryability changes.

**Recommendation:** confirm nothing filtered on it — the collapse is right either way — and if something did, note the replacement in `consumer-queries.md` alongside the other breaking changes in this PR.

PR #355, opened 2026-09-11
fix: 🐛 post-resync defects found testing redesign/07 against real chain data

prashantasdeveloper:
## What

Three fixes found by resync-testing `redesign/07-schema-invariants` from genesis against
testnet — the first full resync of anything past Phase 4, so the first time several code paths
ever ran against real chain data. Branched off `redesign/07-schema-invariants`; base this PR on
it, not `master`.

### 1. `SetAssetMetadataValue` couldn't resolve its key through a batch or multisig proposal

`resolveMetadataKey` only inspected the outer, signed extrinsic. A `setAssetMetadata(Details)`
call reached via `utility.batch*` (the common "create an asset and set its metadata inline"
shape) or executed by `multiSig.approve` of a previously-created proposal never resolved — the
row was silently dropped with a `MissingReferencedEntity` anomaly instead of being written.

Root-caused against 12 real occurrences on testnet (specs `7002000`/`7004000`/`7004001`/
`8000000`). Both new paths pick their call by ordinal position among same-extrinsic
`SetAssetMetadataValue(Details)` siblings (not by `asset_id` alone — a batch/proposal can set more
than one key, including twice for the same asset) and verify against the already-decoded
`assetId` as a consistency guard. `parseMetadataKey` also strips the thousands separators
`toHuman()` puts in a `u64` (`"1,234"`), which would otherwise split a key's rows depending on
which path resolved it. Both of those were caught in an Opus review of the first version of this
fix, along with several test-coverage gaps — see the commit body for the full review trail.

Pre-existing since Phase 5 (`redesign/05-holdings-nfts`, `6149410`); not introduced by phase 7.

### 2. A later v8 runtime names an event's fields in snake_case, crashing the indexer

Found live, mid-resync, as a genuine crash loop (not a candidate item — this blocked the resync
from ever completing). `asset.Approval`'s metadata names its fields in Substrate's idiomatic Rust
snake_case (`asset_id`) on a later v8 testnet spec, not the camelCase (`assetId`) every shape
table and handler in this codebase reads. The decode layer's own doc comment asserted Polymesh
pallets are tuple-style (no field names) at every spec version — true until this runtime — so
there was no shape-table fallback once metadata named a field at all, and nothing caught the
resulting `FieldNotFound` at the handler call site.

Fixed at the root in `metadataFieldNames`, not as a one-event patch: normalise a metadata field
name to camelCase only when it contains an underscore, a no-op for the (previously universal)
already-camelCase case. Deployed directly to the resync server mid-run (code-only change, no
schema drop) to unblock it; confirmed it sailed through the stuck block cleanly.

### 3. A17: lazily-discovered accounts got approximate provenance

`getOrCreateAccount` defaulted `createdEventId` to the block's first event
(`${blockId}/0000000000`) for every account discovered while resolving an asset transfer's
from/to holder — accepted as a bounded phase-7 follow-up rather than an in-phase cascade (see
`docs/implementation/13-entity-provenance.md`).

Threaded the real `blockEventId` through the whole asset-holder resolution chain
(`rawAssetHolderToAssetHolder` → `extractAssetHolder` → `meshAssetHolderToAssetHolder` →
`getOrCreateAccount`) from every real call site: asset transfers/allowances, NFT transfers,
settlement legs and funds transfers. The portfolio-only paths never call `getOrCreateAccount` and
needed no change. A caller with no event to give still gets the old default.

## Testing

- `yarn codegen && yarn typecheck && yarn lint && yarn test:unit && yarn check-handlers && yarn build`
  — all green throughout (final: 0 type errors, 0 lint warnings, **512/512 unit tests**).
- Verified against a full genesis-to-tip testnet resync (`redesign/07-schema-invariants` +
  these fixes): reached chain head cleanly. Final `indexer_anomalies` table has exactly the 12
  rows item 1 explains — nothing else — and **zero** `BalanceReconciliationDrift` rows (better
  than expected; the phase-7 reconcile-hygiene fix left no drift at all across the whole run).

### DIFF
--- a/src/decode/field.ts
+++ b/src/decode/field.ts
@@ -7,15 +7,30 @@ import { FieldNotFound } from './errors';
 export type DecodedEvent = Readonly<Record<string, Codec>>;
 
 /**
+ * `asset_id` -> `assetId`. A no-op for a name with no underscore, so an already-camelCase name
+ * (the common case so far) passes through unchanged rather than being lowercased.
+ */
+const toCamelCase = (name: string): string =>
+  name.includes('_')
+    ? name.replace(/_([a-zA-Z0-9])/g, (_match, c: string) => c.toUpperCase())
+    : name;
+
+/**
  * The field names the block's own metadata gives this event, in parameter order.
  *
  * Since metadata v14 the metadata is self-describing: a struct-style event carries a name per
  * field and a tuple-style one carries none. `mapEvent` already straddles this boundary for
  * `typeName`; this reads the sibling `name`.
+ *
+ * Substrate's own macros name fields idiomatic-Rust snake_case (`asset_id`), while every shape
+ * table and handler in this codebase reads camelCase (`assetId`) - the convention `#[derive]`d
+ * upstream-pallet events (and, on a later runtime, some of Polymesh's own) already happen to
+ * satisfy only when the name has no underscore to begin with. Normalising here, once, is what
+ * lets the shape table and every handler stay camelCase-only.
  */
 export const metadataFieldNames = (event: SubstrateEvent): (string | undefined)[] =>
   (event.event as unknown as GenericEvent).meta.fields.map(({ name }) =>
-    name.isSome ? name.unwrap().toString() : undefined
+    name.isSome ? toCamelCase(name.unwrap().toString()) : undefined
   );
 
 /**
--- a/src/mappings/entities/assets/mapAsset.ts
+++ b/src/mappings/entities/assets/mapAsset.ts
@@ -539,7 +539,7 @@ export const handleAssetTransfer = async (event: SubstrateEvent): Promise<void> 
   let toDid: string;
 
   if (!rawFromHolder.isEmpty) {
-    fromHolder = await rawAssetHolderToAssetHolder(rawFromHolder, block, blockId);
+    fromHolder = await rawAssetHolderToAssetHolder(rawFromHolder, block, blockId, blockEventId);
     fromDid = fromHolder.identityId;
     if (fromDid === emptyDid) {
       return; // We ignore the transfer case when Asset tokens are issued
@@ -549,7 +549,7 @@ export const handleAssetTransfer = async (event: SubstrateEvent): Promise<void> 
   let toHolder: AssetHolderDetails | undefined;
 
   if (!rawToHolder.isEmpty) {
-    toHolder = await rawAssetHolderToAssetHolder(rawToHolder, block, blockId);
+    toHolder = await rawAssetHolderToAssetHolder(rawToHolder, block, blockId, blockEventId);
     toDid = toHolder.identityId;
     if (toDid === emptyDid) {
       toDid = null;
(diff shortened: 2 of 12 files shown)


F-OBrien (commented):
`metadataSiblingOrdinal` and the call lists it indexes into are built by different rules, so they desynchronise. That's one root cause behind three separate wrong-key bugs, which is why I'd suggest reworking the approach rather than patching each.

The ordinal counts every `SetAssetMetadataValue(Details)` event in the extrinsic; `isSetMetadataCall` filters calls named `setAssetMetadata`/`setAssetMetadataDetails`. Zipping those by position only holds if every filtered call emits exactly one counted event, and that fails in both directions:

**1. More events than calls.** `registerAndSetLocalAssetMetadata` emits one of the counted events but isn't in the call filter. Since `metadataKeyFromBatch` is tried before `metadataKeyFromSiblingRegistration` (:237-240), it now intercepts events that used to fall through to the registration path. `batchAll([registerAndSetLocalAssetMetadata(A), setAssetMetadata(A, {Global:5})])` writes the register-and-set value to `A/Global/5` and the other to the registered key — **the two values are swapped**, and the `asset_id` guard passes because it's the same asset. This is a regression against the pre-batch behaviour.

**2. More calls than events.** `BATCH_METHODS` (:114) includes `forceBatch` and `batchOptimistic`, which skip failing calls and continue. A failed earlier `setAssetMetadata` emits nothing, so every later event's ordinal is short by one and the surviving value lands under the failed call's key. `batchAll` is atomic and `batch` halts, so those two are fine.

**3. The fourth path doesn't use the ordinal at all.** `metadataKeyFromSiblingRegistration` (:209) takes the *first* registration event via `findIndex`, so two `registerAndSetLocalAssetMetadata` calls in one batch resolve both setters to the first key — one row clobbered, the other never written. Pre-existing, but only reachable now that batched metadata calls resolve.

Two unrelated bugs in the same file:

**4.** `SetAssetMetadataValue` with `detail = None` clears the lock. On chain `None` means details are *unchanged*, so `detailFrom(null)` returning `isLocked: false` wipes a real lock.

**5.** `handleSetAssetMetadataValueDetails` returns silently when it can't resolve the key — no anomaly, no trace. Same in `handleMetadataValueDeleted` when the key won't parse.

**Suggested direction.** Match structurally instead of positionally: build a normalised tree of the extrinsic's calls and walk it alongside the extrinsic's events. `utility` and `multiSig` emit `ItemCompleted`/`ItemFailed` after each dispatched call, so the event stream already carries the call boundaries — no counting, failed calls are explicit, and anything unrecognised can abort to an anomaly rather than guess a key.

I've prototyped this (`metadataKeyResolver.ts`, handler file drops to ~200 lines) with tests pinning all three cases above and sent the patch via slack.

One thing worth doing regardless of approach: the `codec()` mock in `tests/unit/helpers.ts` has no `toHex()`, so code that builds an id with `toString()` and reads it back with `toHex()` passes every test and fails on real data. Adding `toHex()` to the mock catches that class across all suites.

PR #356, opened 2026-09-14
feat: index corporate actions, checkpoints, ballots and relayer subsidies; merge external-agent entities and retire TransferManager

prashantasdeveloper:
### What's in each commit

1. **`feat: index corporate actions`** — `CorporateAction` + `CorporateActionDefaultConfig`
   entities. `CAInitiated`, `CARemoved`, `RecordDateChanged`, `CALinkedToDoc`,
   `DefaultTargetIdentitiesChanged`, `DefaultWithholdingTaxChanged`, `DidWithholdingTaxChanged`.
2. **`feat: index checkpoints and schedules`** — `Checkpoint` + `CheckpointSchedule` entities.
   `CheckpointCreated`, `ScheduleCreated`, `ScheduleRemoved`. The one real version change in this
   whole domain: `ScheduleCreated`/`ScheduleRemoved` went 3→4 args at chain v6.0.0 (`ScheduleId`
   inserted, payload type changed) — handled as a two-entry decode shape, with a regression test
   asserting both arities decode and a mismatch throws.
3. **`feat: index corporate ballots`** — `CorporateBallot` + `CorporateBallotVote` entities.
   `Created`, `MetaChanged`, `RangeChanged`, `RCVChanged`, `Removed`, `VoteCast` — shape-identical
   across every checked spec version, no version work.
4. **`feat: index the Subsidy/Relayer pallet`** — new `Subsidy` entity. The `relayer` pallet had
   zero indexer coverage despite full SDK support. Renamed paying-key → subsidy at chain v8.0.0;
   both eras are handled, since pre-v8 is the bulk of relayer history on any live chain.
5. **`feat!: merge the TickerExternalAgent entities`** (breaking) — `TickerExternalAgent` →
   `AssetAgent`, `TickerExternalAgentHistory` → `AssetAgentHistory` (these two duplicated their
   write path, which is what actually merges); `TickerExternalAgentAction` → `AssetAgentAction`
   (straight rename, unchanged shape — it answers a different question and is driven from an
   unrelated 20-pallet event table). `AgentGroup` gains the `asset` relation it was missing
   entirely; `AgentGroupMembership.member` is now an `Identity` relation instead of a bare string.
6. **`feat!: retire the TransferManager dual model`** (breaking) — removes the `TransferManager`
   entity, its `Asset.transferManagers` derived field, and its writer. It was documented deprecated
   in its own schema comment yet still written unconditionally; neither consumer queries it.

Commits 1–4 are purely additive — new entities and events that didn't exist before, no existing
query changes. Commits 5–6 are breaking schema changes.

### Consumer impact

**Commits 1–4: none.** New entities, new coverage, nothing existing changes shape.

**Commit 5 needs a coordinated SDK/portal release.** Both consumers query all three renamed
entities:

| Old query | New query | Other changes |
|---|---|---|
| `tickerExternalAgents` | `assetAgents` | `callerId` filters/selections → `identityId` |
| `tickerExternalAgentHistories` | `assetAgentHistories` | `type` is now the `AgentHistoryType` enum (was a plain string) |
| `tickerExternalAgentActions` | `assetAgentActions` | no field changes |

`permissions` on `AssetAgent`/`AssetAgentHistory` stays a JSON-encoded string, not a typed
jsonField — see the docstring on `AssetAgent` in `schema.graphql` for why (the existing
`PermissionsJson` shape was built for a different on-chain permission model and doesn't fit an
`AgentGroup`'s `ExtrinsicPermissions` without a lossy flattening).

**Commit 6: no consumer query changes.** Confirmed neither the SDK nor the portal queries
`transferManagers` today. It's a schema removal, hence the `BREAKING CHANGE` footer, but nothing
downstream needs to change.

### Verification

`yarn codegen && yarn typecheck && yarn lint && yarn test:unit` all pass (571 unit tests, 53
suites), plus `yarn build` (validates every `project.ts` handler name resolves) and the
10-index-per-entity cap check across the whole schema. A fixture test exists for every new
handler. Full genesis resync was not run for this PR (out of scope for the bar here, matching how
earlier phases were verified) — worth doing before merge given how central corporate actions and
checkpoints are to the securities domain.


schema.graphql
```diff
@@ -2313,26 +2317,46 @@ type Investment @entity {
...
+`permissions` stays JSON-in-a-string rather than the `PermissionsJson` jsonField: that type's
+shape (`assets`/`portfolios`/`transactions`/`transactionGroups`, each a flat `{type, values}`) was
+built for the secondary-key permission model. An `AgentGroup`'s `ExtrinsicPermissions` is a
+different, richer on-chain structure (`Whole | These<PalletPermissions> | Except<PalletPermissions>`,
+nesting per-pallet dispatchable names) that does not fit it without a lossy flattening — typing it
+correctly is a separate modeling exercise, deferred rather than forced into the wrong shape here
 """
-type TickerExternalAgent @entity {
-  id: ID! # assetId/callerId
+type AssetAgent @entity @compositeIndexes(fields: [["asset", "identity"]]) {
```
F-OBrien: `AssetAgent` is indexed on both sides, but neither `Asset` nor `Identity` can traverse to it — so "who are this asset's agents" and "which assets is this identity an agent of" both need a separate top-level `assetAgents(filter: …)` query rather than composing into the entity selection.

`@derivedFrom` is virtual: no column, no storage, no write-path cost, and it resolves through the FK indexes already declared right here, so neither entity's 10-index budget is touched. `Asset` already derives seven relations of exactly this shape (`documents`, `holders`, `holdings`, `nfts`, `compliance`, `trustedClaimIssuers`, `mandatoryMediators`), which makes agents the one central relation that can't be reached.

Suggest adding both directions:

```graphql
# Asset
"The identities currently acting as agents for this asset. Past memberships are `AssetAgentHistory`"
agents: [AssetAgent!]! @derivedFrom(field: "asset")

# Identity
"The assets this identity is currently an agent for. Past memberships are `AssetAgentHistory`"
agentOf: [AssetAgent!]! @derivedFrom(field: "identity")
```

The docstrings are worth including: `AssetAgent` holds current membership only, so without them a consumer reasonably reads `asset { agents }` as complete history — the inverse of what `Account.keyAssignments` states explicitly ("current and historical").

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+type DidTax @jsonField {
+  did: String!
+  tax: BigInt!
+}
+
+"""
+A corporate action on an Asset — a dividend, benefit, or issuer notice. Defines the record date,
+target holders and withholding tax that a linked `Distribution` or `CorporateBallot` is subject to
+"""
+type CorporateAction @entity @compositeIndexes(fields: [["asset", "kind"]]) {
```
F-OBrien: `CorporateAction`, `Distribution` and `CorporateBallot` all key on `assetId/localId` — on chain they're the same `CAId`, since a distribution and a ballot are each a CA plus extra data. But only `CorporateBallot` carries `corporateAction: CorporateAction!`. `Distribution` has no equivalent, and `CorporateAction` has no link back to either.

So you can reach a ballot's CA, but not a distribution's, and you can't go from a CA to the benefit it declares — which is the question most consumers of this data will ask first.

Since the ids are already identical, both directions are close to free:

```graphql
# Distribution
corporateAction: CorporateAction!

# CorporateAction
distribution: Distribution @derivedFrom(field: "corporateAction")
ballot: CorporateBallot @derivedFrom(field: "corporateAction")
```

(`Distribution` itself is outside this PR's diff, hence the comment here.)

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+"""
+type BallotVoteEntry @jsonField {
+  power: BigInt!
+  fallback: Int
+}
+
+"""
+One identity's vote in a `CorporateBallot`, from `corporateBallot.VoteCast`
+"""
+type CorporateBallotVote @entity {
```
F-OBrien: This is an append-only log of a value the chain treats as replaceable. `handleBallotVoteCast` creates a row per `VoteCast` keyed on `blockEventId`, but on chain `Votes` is `(CAId, IdentityId) -> Vec<BallotVote>` and voting again while the ballot is open *replaces* the previous vote.

So an identity that changes its mind gets two rows, `ballot.votes` returns both, and nothing marks which is current — a consumer tallying results double-counts them.

This PR already introduces `AssetAgent` / `AssetAgentHistory` for exactly this current-vs-historical split, and ballot votes have the same shape. Either apply that pattern, or key the row on `ballotId/voterId` so it upserts and keep the log separately. At minimum the docstring should say the list is append-only and that the current vote is the most recent row per voter.

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+  kind: CorporateActionKind!
+  declarationDate: Date!
+  "Set once `checkpoint.RecordDateChanged` resolves the record date to an existing checkpoint"
+  recordDate: Date
+  checkpoint: Checkpoint
+  details: String
+  "null = applies to every holder, subject to `targetTreatment`"
+  targetIdentities: [String!]
+  targetTreatment: TargetTreatment
+  defaultWithholdingTax: BigInt
```
F-OBrien: `defaultWithholdingTax` is a `Permill` on chain — parts per million — and it's stored here as a bare `BigInt` with no docstring. Same for `DidTax.tax` and the matching field on `CorporateActionDefaultConfig`.

A consumer reading `10000` has no way to know that means 1%. Please add the unit to all three docstrings; it's the kind of thing that only surfaces once someone has rendered a wrong percentage to an investor.

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+type Subsidy @entity {
+  id: ID! # userKey/payingKey
+  "the subsidised key"
+  beneficiaryAccount: Account! @index
+  "pays the fees"
+  payingAccount: Account! @index
+  "current POLYX limit (remaining)"
+  allowance: BigInt!
+  "cumulative `SubsidyDebited` (v8+ only; 0 for pre-v8 history)"
+  totalDebited: BigInt!
```
F-OBrien: The docstring says "v8+ only; 0 for pre-v8 history", which makes a genuine zero and an unindexable past the same stored value — a consumer can't tell "nothing was ever debited" from "we couldn't know".

Making it nullable, with `null` for the pre-v8 range, keeps the distinction and matches the "unknown is recorded, never guessed" line the review docs take elsewhere.

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+A recurring checkpoint schedule for an Asset, from `checkpoint.ScheduleCreated`. The pre-v6
+period-based model (`period`/`start`/`remaining`) and the v6+ explicit-timestamp model
+(`pendingCheckpoints`) are mutually exclusive on one row, era implicit in which is null
+"""
+type CheckpointSchedule @entity {
+  id: ID! # assetId/scheduleId
+  asset: Asset! @index
+  scheduleId: Int!
+  "v6+ — explicit list of scheduled checkpoint timestamps (`ScheduleCheckpoints.pending`)"
+  pendingCheckpoints: [Date!]
```
F-OBrien: This is captured once from `ScheduleCreated.pending` and never updated, so it goes stale as scheduled checkpoints fire — and `ScheduleRemoved` also carries a `pending` set, which isn't captured either.

That's a reasonable thing to store, but the docstring reads as though it's the live pending set. Suggest saying explicitly that it's the schedule as created, not as it stands now — otherwise a consumer will use it to answer "what's still pending" and be wrong for any schedule that has partly run.

project.ts
```diff
@@ -255,6 +257,18 @@ const filters: Record<string, Record<string, string[]>> = {
...
+    // deprecated from 8.0.0 chain version — superseded by the Subsidy events below
+    AuthorizedPayingKey: ['handleSubsidyApproved'],
+    AcceptedPayingKey: ['handleSubsidyAccepted'],
+    RemovedPayingKey: ['handleSubsidyRemoved'],
+    UpdatedPolyxLimit: ['handlePolyxLimitUpdated'],
+    ApprovedSubsidy: ['handleSubsidyApproved'],
+    AcceptedSubsidy: ['handleSubsidyAccepted'],
+    RemovedPendingSubsidy: ['handleSubsidyRemoved'],
+    RemovedSubsidy: ['handleSubsidyRemoved'],
+    SubsidyDebited: ['handleSubsidyDebited'],
```
F-OBrien: `relayer.RelayedTx(caller, target, result)` is the one event in this pallet with no handler. It's in `EventIdEnum` already (deduplicated into the `## utility ##` section), so it's known, just unregistered.

Is that deliberate scope, or an oversight? The PR models `SubsidyDebited`, which is the fee side of relayed activity, so recording who relayed for whom seems like the natural pair. If it's out of scope, worth a `[]` entry with a comment like the two `chain config, no entity` ones above, so the gap reads as a decision rather than a miss.

schema.graphql
```diff
@@ -2313,26 +2317,46 @@ type Investment @entity {
...
 """
-type TickerExternalAgent @entity {
-  id: ID! # assetId/callerId
+type AssetAgent @entity @compositeIndexes(fields: [["asset", "identity"]]) {
+  id: ID! # assetId/did
   asset: Asset! @index(unique: false)
-  caller: Identity! @index(unique: false)
+  identity: Identity! @index(unique: false)
+  group: AgentGroup
+  permissions: String
```
F-OBrien: Permissions are contained in the Agent Group. Probably better to maintain in one location only.

src/mappings/entities/assets/mapCorporateBallot.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+interface MetaJson {
+  title: string;
+  motions: unknown;
+}
+
+/** `BallotMeta { title: Bytes, motions: Vec<Motion> }` — `title` decodes like any other `Bytes` field */
+const decodeMeta = (rawMeta: Codec): string => {
+  const { title, motions } = rawMeta.toJSON() as unknown as MetaJson;
```
F-OBrien: `decodeMeta` hex-decodes the top-level `title` but passes `motions` through untouched. `BallotMeta.motions` is `Vec<Motion>`, and `Motion { title: Bytes, info_link: Bytes, choices: Vec<Bytes> }` — so every motion title, info link and choice is stored as raw `0x…` hex inside the JSON.

That's the actual content of a ballot, and it comes out unreadable for any consumer. The same decoding needs to walk into each motion and its `choices`, not just the outer title.

src/decode/shapes/corporateActions.ts
```diff
@@ -0,0 +1,57 @@
+import { LAST_V5, V6 } from './consts';
+import { registerShape, stable } from './registry';
+
+/**
+ * `corporateAction` / `checkpoint` / `corporateBallot` pallet parameter shapes.
+ *
```
F-OBrien: These shapes aren't covered by the arity contract test. `CAPTURED_MODULES` in `scripts/sync-metadata.ts:36` is still `['asset', 'balances', 'externalAgents', 'identity', 'settlement', 'staking']`, so none of `corporateAction`, `corporateBallot`, `checkpoint` or `relayer` is captured.

The docstring here says these were walked against the Rust source by hand for five tags, which is exactly the work the contract test exists to stop anyone repeating. Without the pallets in `CAPTURED_MODULES`, an arity change in a future runtime passes CI and surfaces as a worker crash mid-index. Please add the four pallets and regenerate the fixtures.

project.ts
```diff
@@ -326,10 +340,12 @@ const filters: Record<string, Record<string, string[]>> = {
...
     StatTypesAdded: ['handleStatTypeAdded'],
     StatTypesRemoved: ['handleStatTypeRemoved'],
-    TransferManagerAdded: ['handleTransferManagerAdded', 'handleStatisticTransferManagerAdded'],
-    TransferManagerRemoved: ['handleTransferManagerRemoved'],
-    ExemptionsAdded: ['handleExemptionsAdded', 'handleTransferManagerExemptionsAdded'],
-    ExemptionsRemoved: ['handleExemptionsRemoved', 'handleTransferManagerExemptionsRemoved'],
+    // TransferManager (deprecated, retired) is gone; these still feed StatType /
+    // TransferComplianceExemption for the pre-v5 percentage/count restriction model
+    TransferManagerAdded: ['handleStatisticTransferManagerAdded'],
+    TransferManagerRemoved: [],
```
F-OBrien: Retiring `TransferManager` leaves the pre-v5 `Count` restriction with nowhere to go. `handleStatisticTransferManagerAdded` (`mapStatistics.ts:338`) only writes a `StatType` when the type is `Percentage`; a `Count` manager writes nothing at all.

That's inconsistent with the exemption path twenty lines below it, which *does* map `Count` to `StatOpTypeEnum.Count` — so a pre-v5 asset can end up with a `TransferComplianceExemption` referencing a stat type that has no `StatType` row.

Separately, `TransferManagerRemoved: []` above is unregistered, so even the `Percentage` row is never removed when the restriction is lifted.

src/decode/shapes/relayer.ts
```diff
@@ -0,0 +1,43 @@
...
+]);
+
+registerShapes(
+  'relayer',
+  ['ApprovedSubsidy', 'AcceptedSubsidy'],
+  introducedAt(V8, ['userKey', 'payingKey', 'initialPolyxLimit'])
+);
+registerShapes(
+  'relayer',
+  ['RemovedSubsidy', 'RemovedPendingSubsidy', 'SubsidyDebited'],
```
F-OBrien: Two of these three field names don't match the runtime. Mainnet metadata gives:

- `RemovedSubsidy(user_key, paying_key, remaining)`
- `RemovedPendingSubsidy(user_key, paying_key, initial_polyx_limit)`
- `SubsidyDebited(user_key, paying_key, amount)`

Only `SubsidyDebited` is `amount`. It's inert today — the handlers read just `userKey`/`payingKey`, and these v8 events are struct-style so they decode by name from metadata regardless — but the shape table is the record of what an event looks like, and the `tupleEvent` unit tests currently assert the wrong names.

Worth noting `RemovedSubsidy.remaining` is genuinely useful data (the allowance left when the subsidy ended) and isn't captured anywhere.

schema.graphql
```diff
@@ -2313,26 +2317,46 @@ type Investment @entity {
...
+correctly is a separate modeling exercise, deferred rather than forced into the wrong shape here
 """
-type TickerExternalAgent @entity {
-  id: ID! # assetId/callerId
+type AssetAgent @entity @compositeIndexes(fields: [["asset", "identity"]]) {
+  id: ID! # assetId/did
   asset: Asset! @index(unique: false)
-  caller: Identity! @index(unique: false)
+  identity: Identity! @index(unique: false)
+  group: AgentGroup
```
F-OBrien: `group` and `permissions` are never written. `handleExternalAgentAdded` deliberately leaves both unset, and the comment at `mapExternalAgent.ts:14` explains why — `GroupChanged` has no counterpart writing back to `AssetAgent`, so populating them on add alone would go stale on the first group move.

That reasoning is sound, but it lives in the handler, not here. A consumer reading the schema sees two populated-looking fields and has no way to know they're always null in this release. Either say so in the docstrings and point at `AssetAgentHistory` as the real source, or drop the fields until both paths are wired.

schema.graphql
```diff
@@ -2313,26 +2317,46 @@ type Investment @entity {
...
   createdEvent: Event!
 }
 
 """
-Represents a change in the status of an external agent.
+Represents all possible kinds of `AssetAgentHistory` row
+
+"AgentAdded" and "AgentPermissionsChanged" carry `permissions`; `AgentRemoved` and `GroupChanged` may not
+"""
+enum AgentHistoryType {
```
F-OBrien: Worth calling out in the migration notes: `handleGroupChanged` previously wrote `AgentPermissionsChanged` and now writes the new `GroupChanged` member.

The new value is more accurate — a group move isn't a permissions edit — but a group move *does* change an agent's effective permissions, so anything rebuilding a permission timeline now has to treat both members, not just `AgentPermissionsChanged`. Since this release already needs a coordinated SDK/portal rollout, it's cheap to state and expensive to discover later.

tests/entities/assetAgentActions.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  expect(subquery?.errors).toBeFalsy();
+
+  return subquery;
+};
+
+describe('assetAgentActions', () => {
+  it('should return the transactions for ticker', async () => {
+    const subquery = await runActionsQuery('', { orderBy: true, totalCount: true });
+
+    expect(subquery?.data).toMatchSnapshot();
```
F-OBrien: The four snapshot files under `tests/entities/__snapshots__/` were deleted in this PR, but these `toMatchSnapshot()` calls remain. With no committed snapshot Jest writes one rather than comparing, so these assertions pass whatever the data is.

This is the suite that would evidence the "same data, new names" claim behind the entity rename, so it's the worst place to lose the comparison. Please regenerate and commit the snapshots.

src/mappings/entities/assets/mapCheckpoint.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const assetId = await getAssetId(rawAssetId, block);
+  const checkpointId = getNumberValue(rawCheckpointId);
+
+  await Checkpoint.create({
+    id: `${assetId}/${checkpointId}`,
+    assetId,
+    checkpointId,
+    totalSupply: getBigIntValue(rawTotalSupply),
+    datetime: getDateValue(moment) ?? block.timestamp,
```
F-OBrien: The `block.timestamp` fallback here is just wrong. It should'nt be needed as the checkpoint timestamp moment should always be present and typically won't match the block the event actually fires in.

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+linked `CorporateAction`'s distribution amounts verifiable
+"""
+type Checkpoint @entity @compositeIndexes(fields: [["asset", "checkpointId"]]) {
+  id: ID! # assetId/checkpointId
+  asset: Asset! @index
+  checkpointId: Int!
+  totalSupply: BigInt!
+  datetime: Date!
+  "Set when created by a `CheckpointSchedule` rather than manually"
+  schedule: CheckpointSchedule
```
F-OBrien: This is never set. The only `Checkpoint.create` (`mapCheckpoint.ts:73`) doesn't pass `scheduleId`, and nothing writes it afterwards — the `CheckpointSchedule.get` further down is in `handleScheduleRemoved`, setting `removedEvent`. So `schedule` is null on every checkpoint in this release.

The handler comment says linking would mean "scanning every schedule for this asset for a matching pending timestamp". Two cheaper routes:

**1. Authoritative, one chain read.** The checkpoint pallet stores `SchedulePoints: (AssetId, ScheduleId) -> Vec<CheckpointId>` — the chain records exactly which checkpoints each schedule produced. No matching heuristic, no ambiguity.

**2. Index-only, no chain read.** `CheckpointSchedule.pendingCheckpoints` already holds the scheduled moments and `CheckpointCreated` carries the checkpoint's own `moment`, so it's a match against the pending timestamps *for that one asset* — a handful of rows, not every schedule. `CheckpointCreated`'s `Option<IdentityId>` is `None` only when a schedule triggered it, so it also tells you up front whether to look. Weaker than (1) only if two schedules on the same asset share a pending moment.

Also worth noting: that `Option<IdentityId>` distinguishes manual from scheduled for free, and it's currently discarded. As it stands `schedule: null` means "we didn't look", not "manual", so the field carries no information at all.

Either wire it — (2) is cheap and needs no chain read — or drop it until it is. A declared relation that's permanently null is the worst of the three in a breaking release that consumers have to migrate to.

src/mappings/entities/externalAgents/mapExternalAgentAction.ts
```diff
@@ -27,7 +27,7 @@ export async function mapExternalAgentAction(event: SubstrateEvent): Promise<voi
     extrinsic
   );
   if (assetId) {
-    await TickerExternalAgentAction.create({
+    await AssetAgentAction.create({
```
F-OBrien: Commenting here because it's the only part of this file in the diff, but this is about the `ExternalAgentEventsManager` table below that decides what reaches `AssetAgentAction`.

**The table looks out of date and I think it needs a deliberate pass against the runtime rather than a patch.** Its own docstring says it covers *"`statistics`, `sto`, `complianceManager`, `nft` and the rest"* — but there are no `nft` entries at all.

Gaps I could confirm against mainnet metadata (spec `8000020`):

- **The whole `nft` pallet.** `create_nft_collection`, `issue_nft` and `redeem_nft` are the direct counterparts of `asset.create`/`issue`/`redeem`, and the asset ones *are* in the table. So an agent minting a fungible token is recorded and an agent minting an NFT isn't. (`NFTHoldingsUpdated` should probably stay out for the same reason `asset.Transfer` is excluded — it fires on every transfer.)
- **`statistics.SetAssetTransferCompliance`** — the one event in that pallet not covered; the other nine are.
- **`capitaldistribution.Reclaimed`** — `reclaim` is agent-only. Meanwhile `BenefitClaimed` *is* covered although it's emitted by both `push_benefit` (agent) and `claim` (any holder), so it's over-inclusive in exactly the way `asset.Transfer` was excluded for. The agent-only event is missing and the mixed one is in.

Two that look like they may not belong: `asset.PreApprovedAsset` / `RemovePreApprovedAsset` are receiver-side declarations rather than agent-permissioned calls, as is `AssetAffirmationExemption`. Worth confirming they call `ensure_agent_permissioned`, since the class docstring makes that the definition.

**Two things that make this more than a one-off fix:**

1. **Newer runtimes.** There are freeze/frozen events added in 8.1.1/8.1.2 that are almost certainly missing — mainnet is still on `8000020` so they can't be checked from there, but they need to be in before those runtimes are indexed.
2. **Every version replayed from genesis, not just the latest.** The index replays from block 1, so deprecated and renamed events count too — `IssuedNFT` and `RedeemedNFT` were deprecated at 6.0, and `NFTPortfolioUpdated` was renamed to `NFTHoldingsUpdated` at 8.0. A table built only against current metadata silently loses the historical half.

Given both, a mechanised check would be worth more than another manual sweep — something that diffs this table against the runtime's agent-permissioned extrinsics per spec version, in the spirit of what `CAPTURED_MODULES` does for arity. Otherwise the next pallet gets missed the same way `nft` was.

src/mappings/entities/assets/mapCorporateAction.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export const handleCaRemoved = async (event: SubstrateEvent): Promise<void> => {
+  const { block, blockEventId } = extractArgs(event);
+  const { caId: rawCaId } = decodeEvent(event);
+
+  const { localId, assetId } = await getCaIdValue(rawCaId, block);
+
+  const corporateAction = await CorporateAction.get(caId(assetId, localId));
+
+  if (!corporateAction) {
```
F-OBrien: This swallows a "shouldn't happen" case. A `CARemoved` for a CA that was never indexed means either the `CAInitiated` was missed or the id derivation disagrees between the two handlers — both worth knowing about, and neither leaves a trace today.

Same pattern at `:95` and `:113` here, all four ballot handlers in `mapCorporateBallot.ts` (`:61`, `:78`, `:98`, `:115`), and `handleScheduleRemoved` in `mapCheckpoint.ts:118`. Eight silent returns in total.

`mapSubsidy.ts` in this same PR already does it right — `getSubsidyOrAnomaly` records `AnomalyKind.MissingReferencedEntity` with the id it looked for, then returns undefined. Worth lifting that shape into a shared helper and using it for all eight.

src/mappings/entities/assets/mapCorporateAction.ts
```diff
@@ -0,0 +1,193 @@
...
+    kind: toEnum(CorporateActionKind, ca.kind, CorporateActionKind.Other, {
+      enumName: 'CorporateActionKind',
+      block,
+      eventIdx,
+    }),
+    declarationDate: new Date(ca.declDate),
+    recordDate: ca.recordDate ? new Date(ca.recordDate.date) : undefined,
+    details: bytesToString(rawDetails),
+    targetIdentities: ca.targets.identities,
+    targetTreatment: toEnum(TargetTreatment, ca.targets.treatment, TargetTreatment.Exclude, {
```
F-OBrien: The fallback here is `Exclude`, which is a meaningful value rather than a catch-all — and `targetTreatment` is nullable in the schema, so `undefined` is available.

`Include` and `Exclude` are opposites: one targets the listed identities, the other targets everyone *but* them. If a future runtime adds a variant, guessing `Exclude` silently inverts who a corporate action applies to. `toEnum` records an `UnknownEnumValue` anomaly, so it wouldn't be invisible, but the stored row would still read as a confident answer.

Contrast the `CorporateActionKind` call just above, which falls back to `Other` — a genuine catch-all member, so nothing is misrepresented. Suggest leaving `targetTreatment` undefined on an unrecognised value instead of defaulting to one of the two real ones.

schema.graphql
```diff
@@ -2641,6 +2666,170 @@ type DistributionPayment @entity {
...
+"""
+type CorporateAction @entity @compositeIndexes(fields: [["asset", "kind"]]) {
+  id: ID! # assetId/localId
+  asset: Asset! @index
+  localId: Int!
+  kind: CorporateActionKind!
+  declarationDate: Date!
+  "Set once `checkpoint.RecordDateChanged` resolves the record date to an existing checkpoint"
+  recordDate: Date
+  checkpoint: Checkpoint
```
F-OBrien: Never set — nothing in `mapCorporateAction.ts` assigns `checkpointId`, on any path including `handleRecordDateChanged`.

That makes three always-null relations added by this PR: this one, `AssetAgent.group` / `.permissions`, and `Checkpoint.schedule`. Each has a sound reason individually, but together they're a pattern worth deciding on deliberately rather than per-field, because consumers have to migrate to this release and can't tell a null that means "not applicable" from one that means "not implemented yet".

The chain's `RecordDate` is `{ date, checkpoint: CACheckpoint }` where `CACheckpoint` is `Scheduled(ScheduleId, u64)` or `Existing(CheckpointId)`. The `Existing` case carries the checkpoint id directly, so at least that half is resolvable from the event already being decoded here — the `Scheduled` case is the one that needs the checkpoint to be created first.

